PR #3, opened 2015-01-15
Move streamTest and qMetaType registration from DisplayCluster

dnachbaur:


tests/mock/MockNetworkListener.h
```diff
@@ -59,7 +59,7 @@ class MockNetworkListener : public QTcpServer
     DEFLECT_API virtual ~MockNetworkListener();
 
 protected:
-    void incomingConnection(int handle) override;
+    void incomingConnection(int handle);
```
rdumusc: why remove override..?

dnachbaur: To fix Bamboo gcc 4.4 build

rdumusc: We had a fix for that in DisplayCluster, config.h.in defines override to null, maybe we should add it to deflect too?

dnachbaur: Done

rdumusc: yeah! just re-add final / override here :-)
I remember now that we had troubles in DisplayCluster with  "#define final" because some external function somewhere was called final(), which the preprocessor translated to (). But if it works here, all good!

dnachbaur: Oh, I missed this file. My bad. I slightly remember the 'final' problem, but so far it works...

rdumusc:
+1

PR #12, opened 2015-03-26
Include is unused and causes C++11 error when compiled with some 3rd par...

favreau:
Include is unused and causes C++11 error when compiled with some 3rd party applications

### DIFF
--- a/deflect/Event.h
+++ b/deflect/Event.h
@@ -40,7 +40,9 @@
 #ifndef DEFLECT_EVENT_H
 #define DEFLECT_EVENT_H
 
-#include <cstdint>
+#ifdef _MSC_VER
+#  include <stdint.h>
+#endif
 
 #include <deflect/api.h>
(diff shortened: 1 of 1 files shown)


eile:
Might be needed on Windows or OS X?

eile:
Works on OS X.

rdumusc:
It's possible that the include was needed for pre-C++11 compilers, which didn't know the type uint32_t. Don't know about windows.

favreau:
Windows need either #include <stdint.h> or something like:
#ifdef WIN32
typedef unsigned int uint32_t;
#endif
Which one would you prefer?

eile:
```
#ifdef _MSC_VER
#  include <...>
#endif
```

rdumusc:
+1 for this, removal also needed on OS X Lion. Let's merge as soon as Jenkins is back online.

rdumusc:
retest this please

PR #15, opened 2015-04-28
Implement wall->desktop events in DesktopStreamer

eile:


rdumusc:
I don't mind, but is it really cleaner? If you think so, go for it and also please move up the source of SimpleStreamer main.cpp as well :-)

eile:
done

eile:
It's cleaner because it removes an unnecessary directory, imo.

rdumusc:
+1

apps/DesktopStreamer/CMakeLists.txt
```diff
@@ -28,8 +28,9 @@ if(APPLE)
   set(DESKTOPSTREAMER_ICON_FILE desktopstreamer.icns)
 
   # Required to control AppNap on OSX >= 10.9
```
rdumusc: Update comment, not only for AppNap anymore

eile:
Hold merge until it has been tested on the wall.

rdumusc:
+1 - the only point of detail is to maybe move the injection of the mouse events to a separate class?

eile:
Yes, we should do this once it gets more complex.

eile:
Retest this please

rdumusc:
+1, merge when you're ready

PR #16, opened 2015-05-01
Implement ZeroConf discovery for DesktopStreamer

eile:


apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -304,6 +315,25 @@ void MainWindow::processStreamEvents()
     }
 }
 
+void MainWindow::updateServus()
+{
+#ifdef DEFLECT_USE_LUNCHBOX
+    if( hostnameLineEdit_.text() == DEFAULT_HOST_ADDRESS )
```
rdumusc: How does that work? Browsing will stop almost immediately, no?

apps/DesktopStreamer/MainWindow.h
```diff
@@ -106,6 +110,7 @@ private slots:
     QImage cursor_;
 
     QTimer updateTimer_;
+    QTimer browseTimer_;
```
rdumusc: move inside the lunchbox::Servus ifdef?

apps/DesktopStreamer/MainWindow.h
```diff
@@ -70,6 +73,7 @@ private slots:
     void showDesktopSelectionWindow(bool set);
 
     void update();
+    void updateServus();
```
rdumusc, line 77: and I would also have ifdef'd this one...

deflect/NetworkListener.cpp
```diff
@@ -40,48 +40,72 @@
...
-    : pixelStreamDispatcher_(new PixelStreamDispatcher())
-    , commandHandler_(new CommandHandler())
+namespace detail
+{
+class NetworkListener
+{
+public:
+    NetworkListener()
+#ifdef DEFLECT_USE_LUNCHBOX
+        : servus( "_displaycluster._tcp" )
```
rdumusc: This string is defined at two places, shouldn't it be a shared static const like defaultPortNumber_ ?

deflect/NetworkListener.h
```diff
@@ -82,12 +83,11 @@ public slots:
     void onEventRegistrationReply( QString uri, bool success );
 
 private:
+    detail::NetworkListener* _impl;
```
rdumusc, line 89: I am totally open to using leading _ for private members, but the convention in this project is to put them as a prefix. Do we want to gradually change it everywhere, or do we respect the existing practice..?

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -304,6 +315,25 @@ void MainWindow::processStreamEvents()
     }
 }
 
+void MainWindow::updateServus()
```
bilgili, line 320: IMHO updateServus is function should be inside ifdef

apps/DesktopStreamer/MainWindow.cpp
eile: Yes, that's the idea right now. Find a host, enter it in the text box, stop. Otherwise zeroconf would constantly overwrite user input.

apps/DesktopStreamer/MainWindow.h
eile: ifdefs in headers are a bad idea. pimpl would be the solution here.

deflect/NetworkListener.h
eile, line 89: Afair we decided on prefix _, changing it per class, not per project.

deflect/NetworkListener.cpp
eile: done.

apps/DesktopStreamer/MainWindow.cpp
eile: Shit, typo. should be !=. Fixed.

eile:
Ping

eile:
Retest this please

bilgili:
If you guys agree on "_" char :) my +2 is just there !

apps/DesktopStreamer/MainWindow.h
rdumusc: I agree that in the general case ifdefs in headers are bad, but this specific header is not meant to be included anywhere... and since you already introduced an ifdef anyway, it is a good idea to group the browserTimer with the servus. Pimpl is good too, of course.

eile: Did I? damn. Then I'll have to do what you say. Hang in there.

eile:
All ifdefs addressed.

eile:
Retest this please

eile:
@tribal-tec: Seems the wipe out workspace happens after the clone, so your change did not help with this force-push bug. :(

dnachbaur:
No, it wipes before the clone. The new SHA-1 is not present though... Could be a github bug/latency??

eile:
ping

eile:
Removed trailing _ from public members in NetworkListener.

eile:
Retest this please

deflect/NetworkListener.h
```diff
@@ -55,14 +56,17 @@ class NetworkListener : public QTcpServer
 
 public:
     /** The default port number used for Stream connections. */
-    DEFLECT_API static const int defaultPortNumber_;
+    DEFLECT_API static const int defaultPortNumber;
```
rdumusc, line 59: And now I realise that defaultPortNumber is also redefined in the Socket class, but it should only be defined in one place... I'm not sure if it should stay here, or be moved in types.h? Same question for the serviceName...?
Also, I think this class should be renamed from deflect::NetworkListener to deflect::Server, but that would take us too far for this code review...

eile, line 59: Honestly...What does it have to do with my original PR/where do we stop here? 

I'm happy fixing things/making the code better, but it's been a week now that I try to get this relatively simple PR merged.

eile:
ping

deflect/NetworkListener.h
bilgili, line 89: +2 for the code review ( but we should be clear "I mean we all agree" about this prefix, postfix issues at a point, because it is creating complexity between styles, where here prefix is breaking coding style on a very big project )

PR #22, opened 2015-06-22
Improved server api class names. Using common code style guidelines.

rdumusc:
This is intended as an atomic change, but it is divided into several commits for making the making the code review easier.


rdumusc:
Tests are now passing, only Ubuntu VM bug makes it red...

dnachbaur:
retest this please

dnachbaur:
+2. Looks good, but too much to check every line :)

apps/DesktopStreamer/DesktopSelectionRectangle.cpp
```diff
@@ -43,86 +43,86 @@
...
+                                                 mouseEvent )
 {
     // item rectangle and event position
     const QRectF r = rect();
     const QPointF eventPos = mouseEvent->pos();
 
     // check to see if user clicked on the resize button
-    if(fabs((r.x()+r.width()) - eventPos.x()) <= CORNER_RESIZE_THRESHHOLD &&
-       fabs((r.y()+r.height()) - eventPos.y()) <= CORNER_RESIZE_THRESHHOLD)
+    if( fabs((r.x() + r.width()) - eventPos.x()) <= CORNER_RESIZE_THRESHHOLD &&
```
bilgili: std::abs, use the std maths function as much as possible. We had issues in the past because of mixed c and std functions.

apps/DesktopStreamer/DesktopSelectionView.h
```diff
@@ -46,16 +46,15 @@ class DesktopSelectionRectangle;
...
 public:
-    DesktopSelectionView(QWidget* parent);
+    DesktopSelectionView( QWidget* parent );
     ~DesktopSelectionView();
 
-    DesktopSelectionRectangle* getDesktopSelectionRectangle();
-
-protected:
-    void resizeEvent(QResizeEvent * event);
+    DesktopSelectionRectangle* getSelectionRectangle();
```
bilgili, line 52: const ?

apps/SimpleStreamer/main.cpp
```diff
@@ -1,34 +1,72 @@
...
 #endif
 
-// DisplayCluster streaming
-#include <deflect/Stream.h>
-
 bool deflectInteraction = false;
 bool deflectCompressImage = true;
 unsigned int deflectCompressionQuality = 75;
-char * deflectHostname = NULL;
+char* deflectHostname = NULL;
```
bilgili, line 60: global variables ? or just default values ?

apps/DesktopStreamer/DesktopSelectionView.h
rdumusc, line 52: Can't be, MainWindow has to call setCoordinates()

apps/SimpleStreamer/main.cpp
rdumusc, line 60: globals indeed, but this is in a simple 1-file GLUT example app, I don't think that there is a need to rewrite that part...

bilgili:
+1

apps/DesktopStreamer/DesktopSelectionRectangle.cpp
rdumusc: OK, done!

PR #25, opened 2015-07-02
Only show StreamEvent checkbox on supported platform(s) (OSX).

rdumusc:
Only register for events if user asks for it, and always consume events once registred.

### DIFF
--- a/doc/RelNotes.md
+++ b/doc/RelNotes.md
@@ -25,6 +25,8 @@ if you find any other issue with this release.
 * Many classes of the Server part have been renamed
 * Various API fixes
 * Project-wide codestyle cleanup
+* DesktopStreamer: Minor fixes for desktop interaction, only show checkbox on
+  supported platforms (OSX)
 
 ## Deflect 0.5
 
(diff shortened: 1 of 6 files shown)


doc/RelNotes.md
```diff
@@ -25,6 +25,7 @@ if you find any other issue with this release.
 * Many classes of the Server part have been renamed
 * Various API fixes
 * Project-wide codestyle cleanup
+* DesktopStreamer: Only allow desktop interaction on supported platforms (OSX)
```
dnachbaur: remove 'Only'?

dnachbaur:
+1 otherwise

doc/RelNotes.md
rdumusc: The functionality was introduced in v0.5 so that would not work, but I changed the comment to:
DesktopStreamer: Minor fixes for desktop interaction, only show checkbox on supported platforms (OSX)

dnachbaur:
Ah, missed that. Sorry

PR #28, opened 2015-07-14
CMake update after Eyescale/CMake#341

dnachbaur:


tests/mock/CMakeLists.txt
```diff
@@ -13,14 +13,16 @@ set(MOCK_MOC_HEADERS
   MockServer.h
 )
 
+qt5_wrap_cpp(MOC_OUTFILES ${MOCK_MOC_HEADERS})
```
rdumusc: MOC step should already be handled by common_library! I don't know why but I couldn't use common_library for this lib in the past, but good if it works now :-)

dnachbaur: Done

rdumusc:
+1

PR #29, opened 2015-07-15
Fix missing COMMON_PROJECT_DOMAIN

dnachbaur:


rdumusc:
+1 - I suppose you're just fighting a status message here? Seems like COMMON_PROJECT_DOMAIN is only used in DoxygenRule...

dnachbaur:
Well, the default goes to org.doxygen, which is definitely not correct for this library :)
And yes, it's also a status message.

PR #32, opened 2015-07-26
Fix unit test execution on MSVC

dnachbaur:


tests/cpp/SocketTests.cpp
```diff
@@ -50,47 +50,33 @@ namespace ut = boost::unit_test;
 
 BOOST_GLOBAL_FIXTURE( MinimalGlobalQtApp );
 
-BOOST_AUTO_TEST_CASE( testSocketConnectionValidWhenReturnedCorrectNetworkProtocolVersion )
+void testSocketConnect( const int32_t versionOffset )
 {
     QThread thread;
-    MockServer server;
-    server.moveToThread( &thread );
+    MockServer* server = new MockServer( NETWORK_PROTOCOL_VERSION + versionOffset );
```
rdumusc, line 56: Any reason for making this a ptr? Looks equivalent to me, but slightly more complex with the added connection...

rdumusc:
+1

tests/cpp/SocketTests.cpp
dnachbaur, line 56: I'm pretty sure for QObject::deleteLater the object cannot be on the stack.
It's not equivalent, as the deleteLater connection ensures that the object is destroyed in the correct thread. Before, the server object was moved to the worker thread, but destroyed in the main thread. Qt on Windows was complaining about that.

rdumusc, line 56: OK, interesting..! The thread is joined at the end of the scope, so I didn't think about this...

PR #40, opened 2015-09-01
Potential crash fix, reduce log verbosity

dnachbaur:


doc/Changelog.md
```diff
@@ -3,8 +3,9 @@ Changelog {#Changelog}
 
 ## Deflect 0.7
 
-### 0.7.0 (19-08-2015)
+### 0.7.0 (git master)
```
rdumusc: 0.7.1 as this is a new patch? And keep the date?

dnachbaur: 0.7.0 was not released imo. There is no tag in the repo.

PR #44, opened 2015-09-21
Send images from correct thread, fix Zeroconf announce

dnachbaur:


dnachbaur:
FYI @eile

rdumusc:
+1, Looks good. But note that when using the async send the socket is still technically written to from a (single) different thread. I suppose that Qt's warnings don't trigger in this case just because it's a boost thread and not a QThread.

dnachbaur:
I didn't get warnings with asyncSend, but I'll doublecheck. Good point

dnachbaur:
If the async sender has no QApplication, there is no warning. For Qt based async senders, we have to look into it again later. Right now, the only async sender is Equalizer-based applications, and there I noticed we have no event handling, so it's not useful right now anyways.

rdumusc:
OK, good for me for now!

PR #46, opened 2015-09-25
Fix #45: Mouse position for DesktopStreamer is rendered correctly on Retina displays

### LINKED ISSUE #45: DesktopStreamer mouse position does not account for Retina displays
The points are rendered incorrectly in the top-left quarter of the image

dnachbaur:


rdumusc:
+1

PR #52, opened 2015-10-20
Move deflectQml to deflect/qt, rename library to DeflectQt

dnachbaur:


deflect/qt/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (c) 2015, EPFL/Blue Brain Project
+#                     Daniel Nachbaur <daniel.nachbaur@epfl.ch>
+
+set(DEFLECTQT_MOC_HEADERS EventReceiver.h QmlStreamerImpl.h)
+set(DEFLECTQT_PUBLIC_HEADERS QmlStreamer.h)
+set(DEFLECTQT_SOURCES EventReceiver.cpp QmlStreamer.cpp QmlStreamerImpl.cpp)
+set(DEFLECTQT_LINK_LIBRARIES
+  PUBLIC Deflect Qt5::Quick PRIVATE Qt5::Qml)
+set(DEFLECTQT_INCLUDE_NAME deflect/qt)
+set(DEFLECTQT_NAMESPACE deflectqt)
```
rdumusc, line 11: is this correct..?

rdumusc:
+1 otherwise

deflect/qt/CMakeLists.txt
dnachbaur, line 11: Unfortunately yes. It is the concatenated namespaces that form the DEFLECTQT_API declaration in this case (for the dll export stuff). If unspecified, it uses PROJECTNAME, which would not be correct as then the macro would declare 'import'. It should be cleaned up in CMake/common; it has almost nothing to do with namespaces, more like the library/target name now.

rdumusc, line 11: OK, yes, should be cleaned in CMake/common soon, this is confusing!

PR #53, opened 2015-10-22
DISCL-300: Add size hints property to stream

dnachbaur:


deflect/SizeHints.h
```diff
@@ -0,0 +1,93 @@
...
+    //@}
+
+    /** @name Preferred size */
+    //@{
+    unsigned int preferredWidth = UNSPECIFIED_SIZE;
+    unsigned int preferredHeight = UNSPECIFIED_SIZE;
+    //@}
+
+    /** @return true if rhs and this are equal for all sizes. */
+    bool operator == ( const SizeHints& rhs ) const
```
rdumusc: According to the following guidelines, the best practice is to make == operator non-member and noexcept:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-c161-use-nonmember-functions-for-symmetric-operators
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-c87-beware-of--on-base-classes

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -293,6 +305,15 @@ bool QmlStreamer::Impl::_setupRootItem()
              [this]() { _updateSizes( QSize( _rootItem->width(),
                                              _rootItem->height( ))); } );
 
+    connect( _quickWindow, &QQuickWindow::minimumWidthChanged,
```
rdumusc, line 298: Since we're moving to C++11, we could also go crazy and replace _sizeHintsChanged() by lambda functions here :-)

tests/cpp/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@
 # Copyright (c) 2013-2015, EPFL/Blue Brain Project
 #                     Daniel Nachbaur <daniel.nachbaur@epfl.ch>
 #                     Raphael Dumusc <raphael.dumusc@epfl.ch>
+#
+# Change this number when adding tests to force a CMake run: 0
```
rdumusc, line 6: I never understood how this is useful.. running cmake yourself is simpler than thinking about changing a random line in this file...

rdumusc:
very good otherwise!

deflect/SizeHints.h
dnachbaur: I don't see any reference to noexcept, but can be added of course.
The guidelines contradict themselves. operator== clearly needs access to the representation, especially in a struct where there are no accessor functions. This also renders the statement "the number of functions that needs to be modified after a change in representation" wrong. Independently where the function lives, changes to the representation needs to be reflected anyways.

I can make them non-member functions if you want, but it would almost be the first time we do it for a class/struct :)

deflect/qt/QmlStreamerImpl.cpp
dnachbaur, line 298: It saves the private function declaration. Even liking lambdas, I would not have thought of that in this case.

deflect/SizeHints.h
rdumusc: Just thought I would mention it since I had read that recently. :-) All their examples are consistenly non-member. I never used noexecept before and it's obviously not so important. In any case the reference to it was here:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#-c86-make--symmetric-with-respect-to-operand-types-and-noexcept

rdumusc: Thinking about it, I don't see the contradiction. In the case of struct, the public members are also the public API; having accessor functions would make no difference.

deflect/qt/QmlStreamerImpl.cpp
rdumusc, line 298: and it specifically sets one member instead of all six everytime

tests/cpp/CMakeLists.txt
dnachbaur, line 6: As a user your are forced to run cmake after updating the source tree, because git pull && make would not get that new file.

rdumusc, line 6: That's correct, but I guess I just don't see this as a problem in practice. Jenkins does clean builds anyway. But fine with me, it's just a comment :-)

dnachbaur, line 6: In practice you don't do clean builds. You pull and make. And if you would do 'make ServerTests', you'll get an error :) Not in this commit though, as I touched another cmake file as well.

dnachbaur:
Merge?

tests/cpp/CMakeLists.txt
rdumusc, line 6: Or you pull, cmake and make, and if you forgot the cmake step the test you're looking for just simply does not exists yet, so you run cmake at this point... ;-)

PR #60, opened 2015-11-12
DISCL-296: DesktopStreamer improvements

dnachbaur:
- UI file
- remove selection rectangle
- editable list of stream hostnames


apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -68,136 +70,44 @@ typedef __int32 int32_t;
...
 #define CURSOR_IMAGE_FILE     ":/cursor.png"
 
 MainWindow::MainWindow()
     : _stream( 0 )
-    , _desktopSelectionWindow( new DesktopSelectionWindow( ))
-    , _x( 0 )
-    , _y( 0 )
-    , _width( 0 )
-    , _height( 0 )
+    , _cursor( QImage( CURSOR_IMAGE_FILE ).scaled( 20, 20, Qt::KeepAspectRatio))
```
rdumusc: #define 20, and what about retina displays?

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -343,43 +231,28 @@ void MainWindow::_processStreamEvents()
...
-#endif
-
-void MainWindow::_shareDesktopUpdate()
-{
     QTime frameTime;
     frameTime.start();
 
-    const QPixmap desktopPixmap =
-        QApplication::primaryScreen()->grabWindow( 0, _x, _y, _width, _height );
+    QPixmap pixmap = QApplication::primaryScreen()->grabWindow( 0 );
```
rdumusc: why not const?

apps/DesktopStreamer/MainWindow.h
```diff
@@ -49,70 +50,32 @@
...
+class MainWindow : public QMainWindow, public Ui::MainWindow
 {
     Q_OBJECT
 
 public:
     MainWindow();
-
-signals:
-    void streaming( bool enabled );
+    ~MainWindow();
```
rdumusc: is it needed to define it explicitly? Implementation is empty

apps/DesktopStreamer/MainWindow.cpp
dnachbaur: dunno, I just moved the code. 20 \* devicePixelRatio()?

rdumusc:
Otherwise +1, code looks fine. Would you mind adding a screenshot of the what the new UI and hostname selection box looks like?

apps/DesktopStreamer/MainWindow.cpp
rdumusc: that should do!

dnachbaur:
desktopstreamer layout

dnachbaur:
Updated

rdumusc:
Very good! I asked Pawel about the new hostnames for the 2 new displaywalls, but no reply yet. I don't know if you want to wait to include them in this commit, otherwise you can merge and we do a 0.9.1 release next week

rdumusc:
Just in case, the answer is:
bbpav05 -> 5th floor
bbpav06 -> 6th floor

dnachbaur:
You should really relax the per-commit-is-release policy :)

dnachbaur:
Added the new hosts

rdumusc:
Haha, there is no per-commit-is-release policy :-) I was just thinking we could do everything in one go, that's all!

PR #61, opened 2015-11-13
DesktopStreamer: Add application streaming on OSX

dnachbaur:


apps/DesktopStreamer/DesktopWindowsModel.mm
```diff
@@ -0,0 +1,336 @@
...
+        for( ;; )
+        {
+            CFArrayRef windowList =
+                CGWindowListCopyWindowInfo( kCGWindowListOptionOnScreenOnly|
+                                            kCGWindowListExcludeDesktopElements,
+                                            kCGNullWindowID );
+            if( _addApplication( app, windowList ))
+                break;
+            // notification is not sync'd with CGWindowListCopyWindowInfo
+            sleep( 1 );
```
rdumusc, line 208: Just wondering here: this is a long sleep inside a begin/endInsertRows. This never caused problems like blocking the UI?

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -241,8 +279,23 @@ void MainWindow::_shareDesktopUpdate()
     QTime frameTime;
     frameTime.start();
 
-    const QPixmap pixmap = QApplication::primaryScreen()->grabWindow( 0 );
-    _windowRect = QRect( 0, 0, pixmap.width(), pixmap.height( ));
+    QPixmap pixmap;
```
rdumusc, line 282: Ah, this is why the const was gone in the prev code review :-)

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -241,8 +279,23 @@ void MainWindow::_shareDesktopUpdate()
...
+#ifdef DEFLECT_USE_QT5MACEXTRAS
+    if( !_windowIndex.isValid( ))
+        return;
+    if( _windowIndex.row() > 0 )
+    {
+        pixmap = _listView->model()->data( _windowIndex,
+                                           Qt::UserRole ).value< QPixmap >();
+        _windowRect = _listView->model()->data( _windowIndex,
+                                                Qt::UserRole+1 ).value< QRect >();
```
rdumusc: Can you use the correct enum directly here instead of UserRole+1?

apps/QmlStreamer/main.cpp
```diff
@@ -1,41 +1,21 @@
...
-/*                                                                   */
-/* The views and conclusions contained in the software and           */
-/* documentation are those of the authors and should not be          */
-/* interpreted as representing official policies, either expressed   */
-/* or implied, of The University of Texas at Austin.                 */
-/*********************************************************************/
+/* Copyright (c) 2015, EPFL/Blue Brain Project
+ *                     Daniel.Nachbaur@epfl.ch
+ *
+ * This file is part of Monsteer <https://github.com/BlueBrain/Deflect>
```
rdumusc: I don't think so...!

rdumusc:
Good, just minor details! (I don't know Cocoa, so I didn't review that part in much details).

apps/DesktopStreamer/DesktopWindowsModel.mm
dnachbaur, line 208: Hard to say. While you're going back to DeskopStreamer after opening a new application, the sleep is over :)

rdumusc:
Code is good, you can merge now if you want.
FYI as I tested I discovered the following limitations:
- Hidden windows (at the time of launching DesktopStreamer) never appear in the list, even after made visible.
- Fullscreen applications (i.e those on their own "Desktop") are not shown in the list
- Only the main virtual desktop can be streamed
- The list of windows is perfectly functional but not yet super nice to navigate, maybe we can try to improve the look and feel later

dnachbaur:
I put your comments here: https://github.com/BlueBrain/Deflect/issues/63

PR #62, opened 2015-11-15
Fix MSVC build

dnachbaur:


deflect/SizeHints.h
```diff
@@ -73,6 +73,10 @@ struct SizeHints
     //@}
 };
 
+#if _MSC_VER
+#  define noexcept
```
rdumusc: Arff, partial C++11 support :-(
Should we add a version check to that so that it eventually gets deprecated, like:
if defined(_MSC_VER) && _MSC_VER < 1800 ?
http://stackoverflow.com/questions/18387640/how-to-deal-with-noexcept-in-visual-studio

dnachbaur: I was building with ninja, and it only reports _MSC_VER w/o any version iirc. The proper fix would be to add it to CMake/common, like we did for override and final.

rdumusc: OK I see, yes your proposed solution is good

dnachbaur:
retest this please

rdumusc:
+1

PR #65, opened 2015-11-17
QmlStreamer: New 'streamname' command line option

rdumusc:
Needed by DisplayCluster to programatically enforce the (potentially unknown)
name of the AppLauncher Qml streamer.


dnachbaur:
+1

PR #68, opened 2015-12-09
Add access to QmlEngine from QmlStreamer

bilgili:
Qt ImageProvider class needs access to engine for providing images. ( Will soon be needed by Folder View in DisplayCluster )


dnachbaur:
+1

PR #70, opened 2015-12-15
Flush TCP socket after send, Expose Stream from QmlStreamer

bilgili:


deflect/qt/QmlStreamer.cpp
```diff
@@ -66,6 +66,11 @@ QQmlEngine* QmlStreamer::getQmlEngine()
     return _impl->getQmlEngine();
 }
 
+Stream *deflect::qt::QmlStreamer::getStream()
```
dnachbaur, line 69:
```
Stream *deflect::qt:: -> Stream*
```

dnachbaur, line 69: return a reference? Should it be really exposed? What are you needing it for?

deflect/StreamPrivate.cpp
```diff
@@ -117,7 +117,9 @@ bool StreamPrivate::finishFrame()
 {
     // Open a window for the PixelStream
     const MessageHeader mh( MESSAGE_TYPE_PIXELSTREAM_FINISH_FRAME, 0, name );
-    return socket.send( mh, QByteArray( ));
+    const bool ret = socket.send( mh, QByteArray( ));
```
dnachbaur, line 120: flush() on each send or really only for finishFrame?

deflect/qt/QmlStreamer.cpp
bilgili, line 69: It is for sending commands from the file browser.

bilgili:
Do not merge it yet. The flush buffer is not yet solving the refresh problem ( although it seemed to fix with the local copy of display cluster )

deflect/StreamPrivate.cpp
bilgili, line 120: According to our discussion with Raphael, it is at the FinishFrame.

dnachbaur:
It seems the same problem we have with streaming from Equalizer applications where the last frame is not displayed on DisplayCluster. I thought it was due to the missing QApplication in EQ apps, but qml streamer has one. I also played with flush etc, but nothing worked...

deflect/StreamPrivate.cpp
eile, line 120: Makes also sense semantically for me.

deflect/qt/QmlStreamer.cpp
eile, line 69: And the file browser sends raw bytes onto a Deflect-owned stream?

dnachbaur, line 69: I would like to see how it's used. From my naive point of view, I would like to send the command directly from QML, hence it should be provided as a function/signal in Qt. Exposing the stream here means wiring the QML action to the sendCommand outside somewhere. But I have not enough knowledge at this point.

deflect/StreamPrivate.cpp
dnachbaur, line 120: I played with the flush as well. Imo it does not make any difference as we call waitForBytesWritten() which flushes internally. The problem of the missing frame must come from somewhere else. As it is apparently not even the lack of a QApplication, I'm out of ideas (other than replacing QTCPSocket with ZMQ or similar...).

deflect/qt/QmlStreamer.cpp
bilgili, line 69: Check:

Launcher is another process. I am not executing items through the same process. If you guys know how to do signalling from another process without using a connection, please let me know.

https://github.com/BlueBrain/DisplayCluster/pull/264/files
FileBrowser.cpp

rdumusc, line 69: The wiring as done there: https://github.com/BlueBrain/DisplayCluster/pull/264/files#diff-759187a34ccce2f9a287e4731e98066dR76 looks reasonable.
The alternative to not exposing the whole Stream object is be to wrap it here in a small class which extends QObject and expose in QML with engine->setContextProperty(...). The class would have just a few selected public methods such as openFile(QString), and any QML streamer could use this method.

bilgili, line 69: Ok then :) if we want to make it further complicated, most reasonable will be having a REST interface in DC and sending REST requests from qml ;) Almost no C++ afterwards.

deflect/StreamPrivate.cpp
bilgili, line 120: I have done tests myself on this one today and I think it has nothing to do with the socket. Somehow I have noticed that the opened windows in DC are sometimes getting the tap in second time. May be related.

deflect/qt/QmlStreamer.cpp
rdumusc, line 69: As a matter of fact, we DO have a REST interface...!
Maybe we should start using that instead of the deflect sendCommand() which is marked private in doxygen...

bilgili, line 69: :) Lets discuss !

bilgili:
So after discussions and after aha moments, we decided do through the web service interface of the DisplayCluster ( in the past Julio used it for text input ).  So, we don't need to open the stream interface in anyways and it will all be wrapped in the Qml.

I am closing the pull request.

PR #75, opened 2016-02-10
Fix MSVC build

dnachbaur:


deflect/config.h
```diff
@@ -45,8 +45,10 @@
 #    define final
 #    define override
 #  endif
-#  ifndef CXX_NOEXCEPT_SUPPORTED
-#    define noexcept
+#  ifdef CXX_NOEXCEPT_SUPPORTED
+#    define NOEXCEPT noexcept
```
rdumusc, line 49: That's not pretty.. I don't understand why, but I guess you had no choice on MSVC?

rdumusc:
OK otherwise

deflect/config.h
dnachbaur, line 49: Turns out it is somewhat illegal to (re)define C++ keywords:
http://stackoverflow.com/questions/9109377/is-it-legal-to-redefine-a-c-keyword

That's what gcc is not telling you, but MSVC gives a warning here. Would not be a problem if MSVC would support 'noexcept', but iirc even the latest version still has no support for it...

rdumusc, line 49: Admittedly, redefining keywords is a bit playing with fire...

PR #76, opened 2016-02-26
Make window list expandable and increase key hole

eile:


apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -141,73 +137,42 @@ MainWindow::MainWindow()
     connect( &_updateTimer, &QTimer::timeout, this, &MainWindow::_update );
 }
 
+MainWindow::~MainWindow() {}
```
dnachbaur, line 140: rm

dnachbaur:
+1

PR #83, opened 2016-03-21
Release preparation

eile:


dnachbaur:
+1 for me
We postpone the desktopstreamer UI beautifications?

rdumusc:
+1 I would have preferred to have them in, but it is getting too late for the release. Since the UI changes will only affect OSX, which has an independent release cycle, I think it is ok to postpone.

Issue #63, opened 2015-11-16
DesktopStreamer: Improvement for OSX application streaming


dnachbaur:
From https://github.com/BlueBrain/Deflect/pull/61:
- [x] Hidden windows (at the time of launching DesktopStreamer) never appear in the list, even after made visible. (addressed in https://github.com/BlueBrain/Deflect/pull/97)
- [ ] Fullscreen applications (i.e those on their own "Desktop") are not shown in the list
- [ ] Only the main virtual desktop can be streamed
- [x] The list of windows is perfectly functional but not yet super nice to navigate, maybe we can try to improve the look and feel later
- [x] If window hidden (or image size() is 0), don't stream anymore. (addressed in https://github.com/BlueBrain/Deflect/pull/97)
- [x] Show mouse cursor only on window below the mouse cursor
- [ ] Added/removed window(s) for an already open app are not shown in the list
- [x] Apps with more than one window don't stream any image (use case: two X11 windows from vglrun)
- [ ] Hiding, or closing, one of multiple windows of an application leaves a window behind in the list. Closing the full application does not even make it disappear. 
- [ ] Clicking in an application window brings it fullscreen, but the click event itself may be lost if the operation is not fast enough.
- [ ] Use deflect::SizeHints to support resize of streamed window(s)
- [ ] Download pixmap in background thread[s] for better performance
- [ ] Only download changed windows
- [ ] Change the UI so it's more intuitive. It's not obvious that there is a 'Stream' button that needs to be clicked after selecting the windows

PR #84, opened 2016-03-30
DesktopStreamer: Grid-view like layout in apps listview

dnachbaur:


dnachbaur:
FYI @eile

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -72,6 +72,9 @@ const std::vector< std::pair< QString, QString > > defaultHosts = {
     { "DisplayWall 6th floor", "bbpav06.epfl.ch" }
 };
 
+const QString streamAll = "Stream all apps";
```
eile: Just "Stream all"?

eile:
Looks nice, but my desktop image is too small: (cry)
screen shot 2016-03-31 at 8 47 09

dnachbaur:
Hmm, dual-desktop setup I guess? I'll check what's happening there.

eile:
Triple. Could be a fu since the main desktop display is not retina, but the builtin display is? For me is a low prio bug if it's working on the builtin display only.

dnachbaur:
It only gets the image from the first desktop, but the size is all desktops. If streamed, the right part is black. So no multi-desktop support right now. It's part of https://github.com/BlueBrain/Deflect/issues/63 already.

dnachbaur:
Looks like this now with 1 desktop
screen shot 2016-03-31 at 5 59 07 pm

eile:
+1 from me, as implicitly said above.

dnachbaur:
Next/last iteration
screen shot 2016-04-01 at 11 13 40 am

eile:
Unselect/Select all is gone?

rdumusc:
Sorry, merged already. @tribal-tec changed the behaviour a bit from the first version of the PR. Both versions work for me. I think this is also OK because there are different ways to select all windows (rectangle selection or cmd-a) and the 'stop streaming' stops all.

PR #88, opened 2016-04-14
CMake/common adaptions

dnachbaur:


CMakeLists.txt
```diff
@@ -3,41 +3,43 @@
...
 
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/CMake
                               ${CMAKE_SOURCE_DIR}/CMake/common)
 include(GitExternal)
 
-set(VERSION_MAJOR 0)
-set(VERSION_MINOR 10)
-set(VERSION_PATCH 1)
-set(VERSION_ABI 3)
+set(Deflect_VERSION_ABI 3)
```
rdumusc: Why is this variable Deflect_  when all the others below are DEFLECT_ ? Not consistent!

CMakeLists.txt
```diff
@@ -3,41 +3,43 @@
...
+set(DEFLECT_DEB_DEPENDS libjpeg-turbo8-dev libturbojpeg freeglut3-dev
+  libboost-test-dev libboost-date-time-dev libboost-program-options-dev
+  libboost-serialization-dev libboost-system-dev libboost-thread-dev
+  qtbase5-dev)
+set(DEFLECT_PORT_DEPENDS boost freeglut qt5)
 
 include(Common)
 
-common_package(Boost REQUIRED COMPONENTS date_time program_options
+common_find_package(Boost REQUIRED COMPONENTS date_time program_options
```
rdumusc, line 26: nice renaming!

deflect/config.h
```diff
@@ -40,16 +40,20 @@
 #ifndef DEFLECT_CONFIG_H
 #define DEFLECT_CONFIG_H
 
-#ifdef __cplusplus
-#  ifndef CXX_FINAL_OVERRIDE_SUPPORTED
+#ifdef DEFLECT_USE_CXX03
```
rdumusc: Nice too. Should I add Deflect to the list of projects in Jenkins which build with CXX03?

CMakeLists.txt
dnachbaur: It's consistent with Deflect_VERSION_MAJOR etc which originates from project() command (it uses PROJECT_NAME as the prefix which is 'Deflect').
Renaming all the other variables we consume is another iteration I'm afraid.

deflect/config.h
dnachbaur: Actually, Deflect never built (I guess) or is not meant to built with C++03. At least it doesn't now as there errors in other places (MTQueue, non-static member initialization etc.). Not sure how useful this file actually is then.

CMakeLists.txt
rdumusc: OK I see, that's fine then. For cosmetics I would place this line right under project(Deflect VERSION 0.10.1)

PR #98, opened 2016-06-02
Stream constructor uses the DEFLECT_ID and DEFLECT_HOST ENV_VARs

rdumusc:


deflect/Stream.h
```diff
@@ -86,18 +86,22 @@ class Stream
...
-     * @param address Address of the target Server instance, can be a
-     *                hostname like "localhost" or an IP in string format like
-     *                "192.168.1.83".
+     * @param id The identifier for the stream. If set, the environment variable
+     *           DEFLECT_ID takes precedence over the given argument.
+     *           If both values are empty, the constructor will throw.
+     * @param host The address of the target Server instance. It can be a
+     *             hostname like "localhost" or an IP in string format like
+     *             "192.168.1.83". If set, the environment variable DEFLECT_HOST
+     *             takes precedence over the given value.
```
eile: Not sure about the precedence. Let's say a UI allows the user to enter host+port to connect to - know the app has now way of ensuring that this user-provided values are used. Imo the typical use case is that env variables provide defaults, overridable by command line args, overridable by GUI (cf. $DISPLAY, '--display' usage).

rdumusc: That's a fair point. @tribal-tec argued the other way around, that ENV variables should be able to override runtime behaviour without recompiling. Maybe command line arguments should still be able to take precedence?

eile: Imo the default (env, argv) construction should have a different ctor. zeroeq::http::Server has the create() factory method which may return a nullptr - arguably not the most elegant way. The alternative proposed by @tribal-tec was to have the ctor throw if no entity is desired, but this requires the callee to catch - also not great.

rdumusc:
Updated with precedence for the programmatic arguments

dnachbaur:
+1 for me

PR #99, opened 2016-06-03
Fix incomplete socket send under certain timing conditions

dnachbaur:
Observed with Stream::asyncSend() and offscreen Equalizer-based application
streaming: the last frame before the applications is idling is not received
by the server, due to missing flush of the send buffer.


rdumusc:
+1, I hope this fixes the issue with the Qml Launcher as well...

dnachbaur:
No, it doesn't... :(

rdumusc:
I've seen this too, but if the sending is fixed this at least narrows down the search for the problem.
Merge now?

PR #101, opened 2016-06-07
DesktopStreamer: windows that are streamed independently are activated

jafyvilla:
(i.e. sent to the foreground) before applying an interaction event.

The mouse cursor is now rendered only on active windows or desktop.
Fix crash when hiding (cmd+H) the DesktopStreamer app.

### DIFF
--- a/apps/DesktopStreamer/DesktopWindowsModel.mm
+++ b/apps/DesktopStreamer/DesktopWindowsModel.mm
@@ -172,18 +172,28 @@ QRect getWindowRect( const CGWindowID windowID )
     parent = parent_;
 }
 
+- (bool)isCurrent:(NSRunningApplication*)app_
+{
+    NSRunningApplication* current = [NSRunningApplication currentApplication];
+    return [app_ processIdentifier] == [current processIdentifier];
+}
+
 - (void)newApplication:(NSNotification*) notification
 {
     NSRunningApplication* app =
             [[notification userInfo] objectForKey:@"NSWorkspaceApplicationKey"];
-    parent->addApplication( app );
+
+    if( ![self isCurrent:app] )
+        parent->addApplication( app );
 }
 
 - (void)closedApplication:(NSNotification*) notification
 {
     NSRunningApplication* app =
             [[notification userInfo] objectForKey:@"NSWorkspaceApplicationKey"];
-    parent->removeApplication( app );
+
+    if( ![self isCurrent:app] )
+        parent->removeApplication( app );
 }
 
 @end
@@ -209,7 +219,8 @@ public:
     {
         APPNAME,
         WINDOWID,
-        WINDOWIMAGE
+        WINDOWIMAGE,
+        PID
     };
 
     void addApplication( NSRunningApplication* app )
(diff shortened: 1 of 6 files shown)


jafyvilla:
Hold on, I got it wrong for Linux builds

jafyvilla:
Updated

apps/DesktopStreamer/DesktopWindowsModel.mm
```diff
@@ -260,8 +271,9 @@ void reloadData()
                                             kCGNullWindowID );
 
         _data.clear();
-        _data.push_back( std::make_tuple( "Desktop", 0,
-            getPreviewPixmap( QApplication::primaryScreen()->grabWindow( 0 ))));
+        QPixmap preview =
```
rdumusc: const

rdumusc:
Having this queue of events is a bit annoying but a suppose there was no other way
+1 for me, looks good

jafyvilla:
It's the best workaround I could find to the fact that the applications take an iteration of the NSRunLoop to become active after the 'activate' call. Which means that the triggering event gets lost if we don't keep it. I store it in a queue, and it sounds annoying, but in practice this queue won't get too long, as normally all the events are consumed almost immediately.

apps/DesktopStreamer/MainWindow.cpp
```diff
@@ -236,7 +239,7 @@ void MainWindow::_updateStreams()
         const QPersistentModelIndex index; // default == use desktop
         StreamPtr stream( new Stream( *this, index,
                                       _streamIdLineEdit->text().toStdString(),
-                                      host ));
+                                      host, 0 ));
```
dnachbaur: make the PID=0 default parameter? This looks awkward in the non-OSX case.

apps/DesktopStreamer/Stream.cpp
```diff
@@ -162,46 +169,66 @@ std::string Stream::update()
...
-    CGEventPost( kCGHIDEventTap, event );
-    CFRelease( event );
+
+#ifdef DEFLECT_USE_QT5MACEXTRAS
+    // If the destination app is not active, store the event in a queue and
+    // consume it after it's been activated (next iteration of main run loop)
+    if( !DesktopWindowsModel::isActive( _pid ))
+    {
+        DesktopWindowsModel::activate( _pid );
+        std::queue< CGEventRef >().swap( _events ); // empty the queue
```
dnachbaur: typedef the event queue? Used in two places.

apps/DesktopStreamer/Stream.h
```diff
@@ -69,22 +69,11 @@ class Stream : public deflect::Stream
...
+    const QPersistentModelIndex& getIndex() const;
 
 private:
-    const MainWindow& _parent;
-    const QPersistentModelIndex _window;
-    QRect _windowRect; // position on host in non-retina coordinates
-    const QImage _cursor;
-    QImage _image;
-
-    Stream( const Stream& ) = delete;
```
dnachbaur, line 81: non-copy and non-moveable still applies, why removed?

jafyvilla:
Done

dnachbaur:
+1

Issue #96, opened 2016-05-30
DesktopStreamer hangs for several seconds in presentations


rdumusc:
Introduced somewhere with the multi-window interaction. Reproduction: stream anything and wait a bit, then try an action (e.g. change the slide of a presentation). The change takes several seconds to be visible on the wall.

eile:
Likely a bug with the update timer being set to idle.

PR #103, opened 2016-06-09
DesktopStreamer: prevent AppNap of being re-enabled automatically

jafyvilla:
Otherwise, it switches back on after some time
(e.g. when the window is occluded)


jafyvilla:
This should fix #96

PR #106, opened 2016-06-14
DesktopStreamer: Rename 'interaction' -> 'remote control', off by default

rdumusc:


dnachbaur:
+1 for me

eile:
Why off by default?

rdumusc:
My arguments for default off are:
- Not needed for just presenting contents (most common use case)
- With the new interaction mode in Tide, "interactive" windows can only be moved/resize by the borders (less user friendly than regular windows)
- interaction with multi-window streaming is still not prefectly solved
- Users may not want anyone controlling their system without their explicit consent
- interaction is limited anyway by the lack of an on-screen keyboard on the wall

eile:
Ok, just be aware that 90% of the users then will never know this feature exists.

eile:
Could be an icon/button on the wall to enable/disable interaction?

jafyvilla:
+1 for disabling it by default, at least for now. Then we can think of a way to enable it selectively for each window, from the wall

rdumusc:
We can think about letting users do it from the wall - but trying to avoid going back to the two-state situation we had before, though. In terms of implementation it would need an addition to the protocol. I feel that if we do it, this should also prompt for validation on the DesktopStreamer side (at least the first time).
Let's merge this for now, and see how users react. Ideas about how to make the feature more visible for users are welcome :-)

PR #107, opened 2016-06-14
Another attempt to fix the server test for CI

dnachbaur:


rdumusc:
+1, let's hope this attempt will fate better

PR #114, opened 2016-07-04
QmlStreamer handles keyboard events

rdumusc:


dnachbaur:
+1. Too bad we have no unit tests for QmlStreamer :)

PR #121, opened 2016-08-11
Attempt to fix problematic unit test

rdumusc:
deflect::Stream::registerForEvents: receive bind reply failed
../Deflect/tests/cpp/ServerTests.cpp(230): error in
"testDataReceivedByServer": check !"reachable" failed
QThread: Destroyed while thread is still running


dnachbaur:
+1

PR #122, opened 2016-08-12
Also send first frame after request to not lose it

dnachbaur:


rdumusc:
+1, works in conjunction with https://github.com/BlueBrain/Tide/pull/77 to never risk loosing the first frame

Issue #125, opened 2016-08-21
Allow overriding defaultHosts


ghost:
It would be nice if by setting a variable during build time one could pass it's own list of hosts instead of having to modify the hardcoded values from `Deflect/apps/DesktopStreamer/MainWindow.cpp`

rdumusc:
Implementeded with a CMake cache variable, see #126

ghost:
that works.

Thanks!

PR #126, opened 2016-08-30
DesktopStreamer: Added DEFLECT_DESKTOPSTREAMER_HOSTS CMake option

rdumusc:


dnachbaur:
Not much better in my opinion, but ini file kind of thing needs more thinking wrt deployment etc. I guess.
+1

rdumusc:
Yes, this way users can automate a custom build with something like:
cmake .. -DDEFLECT_DESKTOPSTREAMER_HOSTS="{\"My Machine\", \"localhost\"}"
Which is more convenient than patching the source code and should answer request #125

nbalonso:
that works.

Thanks!

PR #128, opened 2016-09-13
Transmit pinch & touch points events to clients [DISCL-386]

rdumusc:


rdumusc:
retest this please

eile:
What is the motivation/use case for touch point transmission? Can we document this, and how it interacts with higher level events?

rdumusc:
Right, I will improve documentation. Reasons are:
- Ability to have more than one touch point (e.g. draw with multiple fingers on a whiteboard)
- Simplified integration for applications that natively handle touch events, like Qt-based ones (although this is not super visible in the DeflectQt streamer case because of the mouse hack that I had to keep).

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -341,6 +341,14 @@ void QmlStreamer::Impl::_onWheeled( double x_, double y_, double deltaY )
     QCoreApplication::postEvent( _quickWindow, e );
 }
 
+void QmlStreamer::Impl::_onPinched( double x_, double y_, QPointF delta )
+{
+    const QPoint point( x_ * width(), y_ * height( ));
+    QWheelEvent* e = new QWheelEvent( point, delta.y(), Qt::NoButton,
```
dnachbaur: pinch should not be translated to wheel I thought?!? QmlStreamer should or can handle gestures/touch points natively, no?

rdumusc: you're right, and this is removed in the second commit which introduces the touch points in deflect. I had put it there because it was an independent commit in the beginning, but maybe this was already a mistake. I can amend this commit.

rdumusc:
Updated, comments addressed.
I was working on an extra commit to move the "touchToMouse" webbrowser hacks into a separate class, but I am going to drop it, it doesn't improves the code the way I had hoped it would.

deflect/qt/CMakeLists.txt
```diff
@@ -3,9 +3,21 @@
...
-set(DEFLECTQT_HEADERS EventReceiver.h QmlGestures.h QmlStreamerImpl.h)
-set(DEFLECTQT_PUBLIC_HEADERS QmlStreamer.h)
-set(DEFLECTQT_SOURCES EventReceiver.cpp QmlStreamer.cpp QmlStreamerImpl.cpp)
+set(DEFLECTQT_HEADERS
+  EventReceiver.h
+  QmlGestures.h
+  QmlStreamerImpl.h
+)
+set(DEFLECTQT_PUBLIC_HEADERS
+  TouchInjector.h
```
dnachbaur: alphabetic order?

deflect/Event.h
```diff
@@ -60,6 +60,50 @@ namespace deflect
...
+ * - EVT_TOUCH_UPDATE        < touch point moved (key field contains point id)
+ * - EVT_TOUCH_REMOVE        < touch point removed (key field contains point id)
+ *
+ * Keybord events
+ * - EVT_KEY_PRESS           < key pressed
+ * - EVT_KEY_RELEASE         < key released
+ *
+ * Deprecated/unused:
+ * - EVT_NONE                < normally unused
+ * - EVT_WHEEL               < not sent by Tide >= 1.2, replaced by EVT_PINCH
```
dnachbaur: wheel is not deprecated imo, in a mouse-only deflect server it fits perfectly with the upper move, click, etc.

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -108,17 +110,21 @@ QmlStreamer::Impl::Impl( const QString& qmlFile, const std::string& streamHost,
     , _stream( nullptr )
     , _eventHandler( nullptr )
     , _qmlGestures( new QmlGestures )
+    , _touchInjector( new TouchInjector( *_quickWindow,
+                                         std::bind( &Impl::_mapToScene, this,
```
dnachbaur, line 114: no lambda?

deflect/qt/TouchInjector.cpp
```diff
@@ -0,0 +1,171 @@
...
+    switch( eventType )
+    {
+    case QEvent::TouchBegin:
+        return Qt::TouchPointPressed;
+    case QEvent::TouchUpdate:
+        return Qt::TouchPointMoved;
+    case QEvent::TouchEnd:
+        return Qt::TouchPointReleased;
+    default:
+        throw std::runtime_error( "unexpected QEvent::Type!" );
```
dnachbaur: print eventType?

deflect/Event.h
```diff
@@ -60,6 +60,50 @@ namespace deflect
...
+ *
+ * Stream events
+ * - EVT_VIEW_SIZE_CHANGED   < remote stream window resized by user
+ * - EVT_CLOSE               < remote stream window closed by user
+ *
+ * Basic interaction
+ * - EVT_PRESS               < touch/mouse button press (single touch point)
+ * - EVT_MOVE                < touch point/mouse move (single touch point)
+ * - EVT_RELEASE             < touch/mouse button release (single touch point)
+ * - EVT_CLICK               < tap/click (with one or more fingers/buttons)
```
dnachbaur: add: key field contains number of fingers/buttons (as used in Tide)

deflect/qt/CMakeLists.txt
rdumusc: done

deflect/Event.h
rdumusc: done

deflect/qt/TouchInjector.cpp
rdumusc: this is really never going to happen, but done

deflect/Event.h
rdumusc: done as discussed

deflect/qt/QmlStreamerImpl.cpp
rdumusc: Did you mean doing:
[this](const QPointF& normalizedPos) { return _mapToScene( normalizedPos ); }
instead of
std::bind( &Impl::_mapToScene, this, std::placeholders::_1 );
?
I find the second one cleaner because it automatically deduces the argument type and avoids a repetition (like the new QObject::connect syntax). Now also imagine the function had three parameters...

dnachbaur: True. In C++ 14 you could use auto for the argument. Ok then.

doc/Changelog.md
```diff
@@ -5,6 +5,17 @@ Changelog {#Changelog}
...
+* [128](https://github.com/BlueBrain/Deflect/pull/128)
+  New events for transmitting all touch points in addition to existing gestures:
+  - Gives the ability to handle more than one touch point in applications (e.g.
+    draw with multiple fingers on a whiteboard).
+  - Simplifies integration for applications that natively handle multitouch
+    events, like Qml-based ones.
+  - DeflectQt uses a new TouchInjector class for sending touch events to Qt
+    (moved from Tide).
+  In addition:
+  - The *wheel* event has been replaced by a more aptly named *pinch* event.
```
dnachbaur: No true now. Pinch is just new, but wheel still stays.

rdumusc: done

dnachbaur:
+1

PR #130, opened 2016-09-26
Replace boost with C++11 [DISCL-391]

rdumusc:
Notes:
- FramePtr must remain a boost::shared_ptr as long as Tide is
  serializing it and does not mandate boost >= 1.56.
- boost::signals2 for Stream::disconnected is hard to replace,
  however it is only used internally by QmlStreamerImpl.cpp.
- program_options and unit_test could be optional.


deflect/Stream.h
```diff
@@ -132,7 +126,7 @@ class Stream
     /** @name Asynchronous send API */
     //@{
     /** Future signaling success of asyncSend(). @version 1.1 */
```
dnachbaur: That breaks API imo. No longer version 1.1

dnachbaur (commented):
Hmm, not removing boost entirely makes a bit wonder how useful that change is. But +1.

rdumusc:
Yes, I thought I could easily remove all when I started. But I couldn't find an elegant solution for the shared_ptr serialization. I could copy the Frame in Tide from std::shared_ptr to boost::shared_ptr before serializing, which wouldn't be too expensive because the segments are QByteArray (implicitly shared) but still look a bit awkward. The boost::signal could be replaced by a callback function which would be fine for our use case. I was not sure and decided to stop here, which is still a step in the right direction (the only problem is the one API breakage you pointed out).

rdumusc:
And I notice now that the API change requires a change in Equalizer because eq/deflect/Proxy.cpp because it explicitly uses boost in a make_ready_future() function...

rdumusc:
Updated: I decided that since this was already breaking the API I might as well do all the changes and make boost optional.

CMakeLists.txt
```diff
@@ -17,15 +17,13 @@ set(DEFLECT_LICENSE BSD)
 set(DEFLECT_DEPENDENT_LIBRARIES Boost)
```
dnachbaur: can be removed now?

doc/Changelog.md
```diff
@@ -5,6 +5,9 @@ Changelog {#Changelog}
 
 ### 0.12.0 (git master)
 
+* [130](https://github.com/BlueBrain/Deflect/pull/130)
+  Replaced boost by C++11. Boost is now an optional dependency and it is used
+  only by the tests. Minor API changes were introdued by this change.
```
dnachbaur: Not so minor imo.
typo 'introdued'

CMakeLists.txt
rdumusc: done

doc/Changelog.md
rdumusc: done, replaced minor by some

dnachbaur:
+1

PR #133, opened 2016-10-28
QmlStreamer: Use asynchronous rendering and image streaming; reuse QuickRenderer from Tide

dnachbaur:


deflect/qt/CMakeLists.txt
```diff
@@ -9,11 +9,13 @@ set(DEFLECTQT_HEADERS
   QmlStreamerImpl.h
 )
 set(DEFLECTQT_PUBLIC_HEADERS
+  QuickRenderer.h
```
rdumusc: alphabetical order: u after m (also below)

deflect/qt/QuickRenderer.h
```diff
@@ -0,0 +1,147 @@
...
+class QuickRenderer : public QObject
+{
+    Q_OBJECT
+
+public:
+    /**
+     * After construction, move the object to a dedicated render thread before
+     * calling any other function.
+     *
+     * @param quickWindow the window to render into if not @sa offscreen
```
rdumusc: could be clarified - how is it used when offscreen

deflect/qt/QuickRenderer.h
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * To be called from GUI/main thread to trigger rendering.
+     *
+     * This call is blocking until sync() is done in render thread.
+     */
+    void render();
+
+signals:
+    /** Emitted at the end of render(). Originates from render thread. */
+    void afterRender();
```
rdumusc, line 109: From the changes in Tide, I see that it is necessary to connect this signal to do the swapBuffer when not in offscreen mode, right? Maybe add that to the doc?

deflect/qt/QuickRenderer.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    if( !_initialized )
+        return;
+
+    {
+        QMutexLocker lock( &_mutex );
+
+        _context->makeCurrent( _getSurface( ));
+
+        _ensureFBO();
```
rdumusc: I would find it clearer to do:

```
if( offscreen )
    _ensureFBO();
```

instead of:

```
void QuickRenderer::_ensureFBO()
if( !_offscreen )
    return;
```

deflect/qt/QuickRenderer.h
rdumusc: Humm forget this remark, I'm reading it again and the class description + offscreen parameter description contain enough information.

rdumusc (approved):

PR #134, opened 2016-11-01
DeflectQt: Multithreaded rendering is off by default for Qt version < 5.5

dnachbaur:


CMakeLists.txt
```diff
@@ -39,6 +39,12 @@ common_find_package(Qt5Widgets REQUIRED)
 common_find_package(Servus)
 common_find_package_post()
 
+if(Qt5Quick_VERSION VERSION_LESS 5.5)
```
rdumusc: maybe the option should not even be here for Qt < 5.5, since we know that it does not work.

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -178,9 +180,9 @@ void QmlStreamer::Impl::_initRenderer()
 {
     _updateSizes( QSize( _rootItem->width(), _rootItem->height( )));
 
-    _quickRenderer = new QuickRenderer( *_quickWindow, *_renderControl,
-                                        true );
+    _quickRenderer = new QuickRenderer( *_quickWindow, *_renderControl, true );
 
+#ifdef DEFLECTQT_MULTITHREADED
 #if QT_VERSION >= 0x050500
```
rdumusc: #else here should probably just static_assert(false) instead of printing a warning because it does not work at all

rdumusc (commented):
One limitation of this approach is that TideWall loses the multithreaded rendering, because QuickRenderer will be compiled without it despite the fact that it works for the on-screen case. What do you think?

rdumusc (approved):

PR #137, opened 2016-11-15
Deflect Qt: New OffscreenQuickView class (extracted from QmlStreamerImpl)

rdumusc:


deflect/qt/OffscreenQuickView.cpp
```diff
@@ -0,0 +1,247 @@
+/*********************************************************************/
+/* Copyright (c) 2016, EPFL/Blue Brain Project                       */
+/*                     Raphael Dumusc <raphael.dumusc@epfl.ch>       */
```
dnachbaur, line 4: I did 80% of this file, for shame :)

deflect/qt/OffscreenQuickView.cpp
```diff
@@ -0,0 +1,247 @@
...
+    , _qmlComponent{ new QQmlComponent{ _qmlEngine.get() }}
+{
+    if( !_qmlEngine->incubationController( ))
+        _qmlEngine->setIncubationController( incubationController( ));
+
+    // remote URL to QML components are loaded asynchronously
+    connect( _qmlComponent.get(), &QQmlComponent::statusChanged,
+             this, &OffscreenQuickView::_setupRootItem );
+
+    // Now hook up the signals. For simplicy we don't differentiate between
```
dnachbaur: typo 'simplicy'

deflect/qt/OffscreenQuickView.cpp
```diff
@@ -0,0 +1,247 @@
...
+}
+
+QQmlContext* OffscreenQuickView::getRootContext() const
+{
+    return _qmlEngine->rootContext();
+}
+
+QImage OffscreenQuickView::getImage() const
+{
+    return _quickRenderer->fbo()->toImage();
```
dnachbaur: no check if fbo() is valid?

deflect/qt/OffscreenQuickView.cpp
```diff
@@ -0,0 +1,247 @@
...
+    _rootItem->setParentItem( contentItem( ));
+    resize( _rootItem->width(), _rootItem->height( ));
+
+    // emulate QQuickView::ResizeMode::SizeRootObjectToView
+    connect( this, &QQuickWindow::widthChanged,
+             _rootItem.get(), &QQuickItem::setWidth );
+    connect( this, &QQuickWindow::heightChanged,
+             _rootItem.get(), &QQuickItem::setHeight );
+
+    // also bind view to root object, to allow progamatic resize of the rootItem
```
dnachbaur: typo 'progamatic'

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -78,136 +71,47 @@ namespace qt
 
 QmlStreamer::Impl::Impl( const QString& qmlFile, const std::string& streamHost,
                          const std::string& streamId )
-    : _renderControl( new QQuickRenderControl )
     // Create a QQuickWindow that is associated with out render control. Note
```
dnachbaur: that comment does not apply here anymore

deflect/qt/OffscreenQuickView.cpp
rdumusc: you did it for my glory, for which I am grateful :-)

deflect/qt/OffscreenQuickView.cpp
rdumusc: done

deflect/qt/OffscreenQuickView.cpp
rdumusc: It has no reason to be invalid at this point and I don't know what to do if it is not... throw?return an empty image?

deflect/qt/OffscreenQuickView.cpp
rdumusc: done

deflect/qt/QmlStreamerImpl.cpp
rdumusc: done

deflect/qt/OffscreenQuickView.cpp
dnachbaur: After constructor or if render target is not FBO this will be nullptr, so at least empty image to not crash. Maybe also add more documentation in the header in which cases this is actually valid.

rdumusc: You're right, and I realize it is also not very safe that this can only be called from a Qt::DirectConnection (while the FBO is bound). A better way would be to get the image internally just afterRender() and always keep a copy, so that class users don't risk making any mistake.

dnachbaur (approved):

PR #140, opened 2016-12-07
DesktopStreamer: use full user name as default stream name

eile:
Just seeing 'tsf-476-wpa-1-249.epfl.ch' right now on the wall. ;)

Ideally I would want the full user name, e.g., 'Stefan Eilemann' and not
'eilemann' as this change does, but getting this information seems way
harder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,6 @@ notifications:
     on_success: never
 language: cpp
 sudo: false
-cache:
-  ccache: true
-  pip: true
-  directories:
-  - /usr/local
-before_cache:
-  - brew cleanup
 os:
   - osx
 env:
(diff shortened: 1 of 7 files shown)


rdumusc:
hmmm yes, your machine has a strange name ;-) mine looks like 'bluebrain065' or  'bluebrain077.epfl.ch' which was slightly better. Your idea is good, and it is possible to get the full username on OSX with NSFullUserName(), I just tried and it seems to work. On linux could be possible with getpwnam(), I will try that as well. Maybe we could even append something to make it look like "Stefan Eileman's laptop" or  "Raphael Dumusc's desktop".

rdumusc:
See improved solution with full user name: https://github.com/BlueBrain/Deflect/pull/141

rdumusc:
replaced by #142

PR #141, opened 2016-12-07
DesktopStreamer: use full user name as default stream name

rdumusc:


apps/DesktopStreamer/nameUtils.h
```diff
@@ -0,0 +1,39 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef NAMEUTILS_H
+#define NAMEUTILS_H
+
+#include <QString>
+
+/**
+ * Cross-plateform utilities to retrieve the name of the user.
```
dnachbaur, line 26: typo plateform

apps/DesktopStreamer/nameUtils.h
```diff
@@ -0,0 +1,39 @@
...
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef NAMEUTILS_H
```
dnachbaur, line 20: DEFLECT_NAMEUTILS_H or similar prefix

dnachbaur (commented):
+1 otherwise

apps/DesktopStreamer/nameUtils.h
rdumusc: argghh, I'll never get this one right..! thanks

eile:
Daniel's and my concerns addressed and pushed into #140

apps/DesktopStreamer/nameUtils.h
rdumusc: the other files in DesktopStreamer don't do it. Should we use DESKTOPSTREAMER_ everywhere to avoid conflicts (e.g. in STREAM_H ) ?

dnachbaur: NAMEUTILS_H sounds very likely to exist somewhere in the world already as a system header. Prefix with DESKTOPSTREAMER_ seems reasonable too. Or pragma once, what Stefan did.

rdumusc: I added DESKTOPSTREAMER_ everywhere

rdumusc:
@eile oops, I didn't mean for this to become a PR battle :-) In the meantime I noticed a problem in DesktopStreamer's CMakeLists (linking with "-framework Foundation" even if Qt5::MacExtras is not found) and fixed Daniel's comments, but I don't know what your concerns were. I will try to do a diff with you branch and incorporate them.

eile:
haha, same here. Thought this way is faster.

eile:
I opened my PR #140 for editing from you, in case that helps.

eile:
Diff at: https://github.com/BlueBrain/Deflect/pull/140/commits/97dd60ebbcf31fb55cc4eb5d18a2026d8fd5ec85

rdumusc:
OK, in the end I integrated your changes into my branch offline, but github won't let me reopen my PR because I force pushed in the meantime, so if you don't mind I'm opening yet another one.

Note: from what I had found I though USERNAME was Windows and USER was Linux, but it can't hurt to have both.

PR #145, opened 2017-01-24
DesktopStreamer: new slider to adjust the JPEG quality in advanced settings

rdumusc:


eile (approved):
We should separate DesktopStreamer from Deflect, I guess. Strictly this is a new feature, so version should be 0.13.0.

rdumusc:
Yes, versionning of applications vs. libraries is somewhat problematic, and it's especially true in Deflect. Historically we were using PATCH for minor features and MINOR for larger features, with no clearly defined rules it seems [see releases](https://github.com/BlueBrain/Deflect/releases), so I kept doing the same instead of bumping to 1.13.0...

PR #149, opened 2017-02-08
DesktopStreamer: Hide cursor when inactive

eile:


eile:
Retest this please.

apps/DesktopStreamer/Stream.cpp
```diff
@@ -56,21 +56,25 @@
...
-    : _stream( stream )
-    , _parent( parent )
-    , _window( window )
-    , _cursor( QImage( CURSOR_IMAGE_FILE ).scaled(
-                   CURSOR_IMAGE_SIZE * parent.devicePixelRatio(),
-                   CURSOR_IMAGE_SIZE * parent.devicePixelRatio(),
-                   Qt::KeepAspectRatio ))
-    , _pid( pid )
-    {}
+Impl( Stream& stream, const MainWindow& parent,
```
rdumusc, line 64: indentation was correct before? The rest of this class is mostly wrong though...

rdumusc (approved):

apps/DesktopStreamer/Stream.cpp
eile: All other methods are this weirdly indented, so I just kept consistency.

PR #153, opened 2017-03-07
Deflect stereo streaming API refactoring

eile:
FYI @rdumusc @tribal-tec: This is WIP and not fully tested yet. Just opening the PR in case you want to comment already.

### DIFF
--- a/deflect/ReceiveBuffer.h
+++ b/deflect/ReceiveBuffer.h
@@ -47,8 +47,8 @@
 
 #include <QSize>
 
-#include <array>
 #include <map>
+#include <queue>
 
 namespace deflect
 {
@@ -65,8 +65,8 @@ public:
      * Add a source of segments.
      * @param sourceIndex Unique source identifier
      * @return false if the source was already added or if
-     *         finishFrameForSource() has already been called for all existing
-     *         sources (TODO DISCL-241).
+     * finishFrameForSource()
+     *         has already been called for all existing source (TODO DISCL-241).
      */
     DEFLECT_API bool addSource(size_t sourceIndex);
 
@@ -80,13 +80,11 @@ public:
     DEFLECT_API size_t getSourceCount() const;
 
     /**
-     * Insert a segement for the current frame and source.
+     * Insert a segment for the current frame and source.
      * @param segment The segment to insert
      * @param sourceIndex Unique source identifier
-     * @param view in which the segment should be inserted
      */
-    DEFLECT_API void insert(const Segment& segment, size_t sourceIndex,
-                            View view = View::mono);
+    DEFLECT_API void insert(const Segment& segment, size_t sourceIndex);
 
     /**
      * Call when the source has finished sending segments for the current frame.
@@ -95,36 +93,28 @@ public:
      */
     DEFLECT_API void finishFrameForSource(size_t sourceIndex);
 
-    /** Does the Buffer have a new complete mono frame (from all sources) */
-    DEFLECT_API bool hasCompleteMonoFrame() const;
-
-    /** Does the Buffer have a new complete stereo frame (from all sources) */
-    DEFLECT_API bool hasCompleteStereoFrame() const;
+    /** Does the Buffer have a new complete frame (from all sources) */
+    DEFLECT_API bool hasCompleteFrame() const;
 
     /**
      * Get the finished frame.
      * @return A collection of segments that form a frame
      */
-    DEFLECT_API Segments popFrame(View view = View::mono);
+    DEFLECT_API Segments popFrame();
 
     /** Allow this buffer to be used by the next
      * FrameDispatcher::sendLatestFrame */
-    DEFLECT_API void setAllowedToSend(bool enable, View view);
+    DEFLECT_API void setAllowedToSend(bool enable);
 
     /** @return true if this buffer can be sent by FrameDispatcher */
-    DEFLECT_API bool isAllowedToSend(View view) const;
+    DEFLECT_API bool isAllowedToSend() const;
 
 private:
-    std::map<size_t, SourceBuffer> _sourceBuffers;
+    using SourceBufferMap = std::map<size_t, SourceBuffer>;
 
-    /** The current indices of the mono/left/right frame for this source. */
-    std::array<FrameIndex, 3> _lastFrameComplete = {{0u, 0u, 0u}};
-
-    /** Is the mono/left/right channel allowed to send. */
-    std::array<bool, 3> _allowedToSend = {{false, false, false}};
-
-    FrameIndex _getLastCompleteFrameIndex(View view) const;
-    void _incrementLastFrameComplete(View view);
+    FrameIndex _lastFrameComplete = 0;
+    SourceBufferMap _sourceBuffers;
+    bool _allowedToSend = false;
 };
 }
 
--- a/deflect/ReceiveBuffer.h
+++ b/deflect/ReceiveBuffer.h
@@ -65,8 +65,8 @@ public:
      * Add a source of segments.
      * @param sourceIndex Unique source identifier
      * @return false if the source was already added or if
-     * finishFrameForSource()
-     *         has already been called for all existing source (TODO DISCL-241).
+     *         finishFrameForSource() has already been called for all existing
+     *         source (TODO DISCL-241).
      */
     DEFLECT_API bool addSource(size_t sourceIndex);
 
--- a/deflect/Segment.h
+++ b/deflect/Segment.h
@@ -56,6 +56,8 @@ struct Segment
     /** Parameters of the segment. */
     SegmentParameters parameters;
 
+    View view = View::mono; //!< Eye pass for the segment
+
     /** Image data of the segment. */
     QByteArray imageData;
 
--- a/deflect/SegmentParameters.h
+++ b/deflect/SegmentParameters.h
@@ -82,6 +82,8 @@ struct SegmentParameters
 
     /** Data format of the Segment. @version 1.6 */
     DataType dataType = DataType::jpeg;
+
+    View view = View::mono; //!< Eye pass for the segment
 };
 }
 
--- a/deflect/SegmentParameters.h
+++ b/deflect/SegmentParameters.h
@@ -82,8 +82,6 @@ struct SegmentParameters
 
     /** Data format of the Segment. @version 1.6 */
     DataType dataType = DataType::jpeg;
-
-    View view = View::mono; //!< Eye pass for the segment
 };
 }
 
--- a/deflect/Stream.h
+++ b/deflect/Stream.h
@@ -52,8 +52,6 @@
 #include <memory>
 #include <string>
 
-class Application;
-
 namespace deflect
 {
 class StreamPrivate;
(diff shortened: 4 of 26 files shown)


deflect/SegmentParameters.h
```diff
@@ -69,6 +69,8 @@ struct SegmentParameters
 
     /** Is the image raw pixel data or compressed in jpeg format */
     bool compressed = true;
+
+    View view = View::mono; //!< Eye pass for the segment
```
rdumusc: I had started like this but this breaks compatibility with older clients because the serialization code uses sizeof(SegmentParameters). I don't think a workaround is possible for all use cases without a proper protocol version exchange, which we don't have.

eile:
Should be 90% there now. Please review.

eile:
Btw, DesktopStreamer async streaming helps perf :)

rdumusc:
Ouch, I hadn't noticed that DesktopStreamer was NOT using the async send before...! That must help indeed... Will have a look at the changes.

deflect/ReceiveBuffer.h
```diff
@@ -65,8 +65,8 @@ class ReceiveBuffer
      * Add a source of segments.
      * @param sourceIndex Unique source identifier
      * @return false if the source was already added or if
-     *         finishFrameForSource() has already been called for all existing
-     *         sources (TODO DISCL-241).
+     * finishFrameForSource()
+     *         has already been called for all existing source (TODO DISCL-241).
```
rdumusc: broken indentation

deflect/Segment.h
```diff
@@ -56,6 +56,8 @@ struct Segment
     /** Parameters of the segment. */
     SegmentParameters parameters;
 
+    View view = View::mono; //!< Eye pass for the segment
```
rdumusc, line 59: comment style consistency

deflect/Stream.h
```diff
@@ -124,48 +122,47 @@ class Stream
...
+     *
+     * The send (and the optional compression) and finishFrame() are executed in
+     * a different thread. The result of this operation can be obtained by the
+     * returned future object.
+     *
+     * @param image The image to send. Note that the image is not copied, so the
+     *              referenced must remain valid until the send is finished
+     * @return true if the image data could be sent, false otherwise.
+     * @see send()
+     * @version 1.1
```
rdumusc: @version 1.6 also for send() and finishFrame()

deflect/StreamSendWorker.h
```diff
@@ -62,20 +63,32 @@ class StreamSendWorker
     ~StreamSendWorker();
 
     /** Enqueue an image to be send during the execution of run(). */
-    Stream::Future enqueueImage(const ImageWrapper& image);
+    Stream::Future enqueueImage(const ImageWrapper& image, bool finish);
+
+    Stream::Future enqueueFinish(); //!< Enqueue a finishFrame()
```
rdumusc, line 68: doc style consistency

doc/StereoStreaming.md
```diff
@@ -18,11 +18,22 @@ New view enum in deflect/types.h:
 
     enum class View : std::int8_t { mono, left_eye, right_eye };
 
-On the client side, no changes to the Stream API. The ImageWrapper takes an
-additional View parameter.
+The ImageWrapper takes an additional View parameter.
+
+API changes to the Stream class:
+* asyncSend is deprecated but work as before, but only for mono views
```
rdumusc: work -> works

doc/StereoStreaming.md
```diff
@@ -18,11 +18,22 @@ New view enum in deflect/types.h:
...
+
+The send operation sends one tile of one eye pass to the server. The finishFrame
+operation completes the whole frame for all eye passes, that is, send is
+analogous to an OpenGL draw pass (GL_BACK, GL_BACK_LEFT, GL_BACK_RIGHT), and
+finishFrame is analogous to a swap buffer call, with the extension that all
+client's finish is synchronized by the server.
 
 On the server side, no changes to the Server API (except some cleanups). Each
-Frame dispatched now contains the View information.
+Frame dispatched now contains the View information. A frame is considered
```
rdumusc: Each Frame dispatched now contains the View information. -> no longer true, the frame contains the segments for all views and each Segment has the view information.

doc/StereoStreaming.md
```diff
@@ -18,11 +18,22 @@ New view enum in deflect/types.h:
...
+complete when all connected clients did send a finish.
```
rdumusc: did send -> have sent

doc/StereoStreaming.md
```diff
@@ -31,7 +42,7 @@ payload. This message is silently ignored by older Servers.
 
 ## Examples
 
-Example of a stereo 3D client application using the blocking Stream API:
+Example of a stereo 3D client application using synchronous operations:
```
rdumusc, line 45: this section must be corrected to .wait() or be removed

doc/StereoStreaming.md
```diff
@@ -56,18 +67,12 @@ Example of a stereo 3D client application using the blocking Stream API:
...
 
     deflect::Stream stream( ... );
 
     /** ...synchronize start with other render clients (network barrier)... */
 
-    delfect::Stream::Future leftFuture, rightFuture;
-    leftFuture = deflect::qt::make_ready_future<bool>( true );
-    rightFuture = deflect::qt::make_ready_future<bool>( true );
-
-    ImageData leftData, rightData; // must remain valid until sending completes
```
rdumusc, line 69: why remove this? I think it is important to show the correct usage of the api. Without those variables the renderLoop below does not make sense because the "auto future" goes out of scope.

Furthermore there should be a third future for the deflectStream->finishFrame();

tests/cpp/perf/benchmarkStreamer.cpp
```diff
@@ -240,7 +241,8 @@ class Application
                                            _noiseImage.height(), deflect::RGBA);
         deflectImage.compressionPolicy = deflect::COMPRESSION_OFF;
 
-        return _stream->send(deflectImage) && _stream->finishFrame();
+        return _stream->send(deflectImage).get() &&
```
rdumusc: sendAndFinish().get()

tests/cpp/perf/benchmarkStreamer.cpp
```diff
@@ -251,7 +253,8 @@ class Application
         deflectImage.compressionPolicy = deflect::COMPRESSION_ON;
         deflectImage.compressionQuality = _options.quality;
 
-        return _stream->send(deflectImage) && _stream->finishFrame();
+        return _stream->send(deflectImage).get() &&
```
rdumusc: same

deflect/Segment.h
eile: ?

deflect/ReceiveBuffer.h
eile: fixed

deflect/Stream.h
eile: done

deflect/StreamSendWorker.h
eile: I don't understand what you mean here. It's standard doxygen, and we use it everywhere?

doc/StereoStreaming.md
eile: Fixed

doc/StereoStreaming.md
eile: must they? If executed as is, each call should be synchronous, or am I missing something?

doc/StereoStreaming.md
eile: Ah, I interpreted it differently. Re-adding.

deflect/Segment.h
rdumusc: It's just a style remark but all other 3 fields in this class use the /** ... */ syntax, and this one in the middle has a //!< ... on the side. The //!< syntax is more compact, but then I would expect all fields to follow the same style... but it's not a big deal ;-)

tests/cpp/perf/benchmarkStreamer.cpp
eile: damn, one slipped through.

tests/cpp/perf/benchmarkStreamer.cpp
eile: what? ;)

deflect/StreamSendWorker.h
rdumusc: This is new to me though, I've never used that anywhere ;-) what rule do you apply to choose between //!< and /** ... */, you check if it can fit on one line?  It just does not look very consistent to me when I look at the entire file but I can live with it if you like it :-)

doc/StereoStreaming.md
rdumusc: Actually you made me realize yesterday that std::future destructor does not wait(), execpt in the special case of std::async(std::launch::async). http://en.cppreference.com/w/cpp/thread/future/%7Efuture
So I think this is not the same as the previous synchronous api. In fact, if the sending takes longer than the rendering this will queue up send requests indefinitely, or more likely crash because an ImageWrapper will point to invalid data...

rdumusc:
FYI Tide adaptation: https://github.com/rdumusc/Tide/commit/0e4d1b42e36fc6550e04dd3642a2a0be0417695e

doc/StereoStreaming.md
eile: Gah, you're right. Changing it.

eile:
Ready to squash & merge, as far as I can see.

rdumusc (approved):

eile:
Please review two last commits, RTM from my side.

deflect/StreamSendWorker.cpp
```diff
@@ -68,40 +68,43 @@ void StreamSendWorker::_run()
...
         while (_requests.empty() && _running)
             _condition.wait(lock);
 
         if (!_running)
             break;
 
         const Request& request = _requests.front();
+        const ImageWrapper image(request.image);
+        const uint32_t tasks(request.tasks);
+        PromisePtr promise(request.promise);
```
rdumusc: I find the way the request is copied quite confusing. I didn't see at first that "image" was a copy, I thought it was wrapping around "request.image". I would simply do in two lines:

    const auto request = _requests.front();
    _requests.pop_front();

Or if for some reason you prefer having individual variables then I would still suggest C++11 style with auto for readability:

    const auto& request = _requests.front();
    const auto image = request.image;
    const auto tasks = request.tasks;
    const auto promise = request.promise; 
    _requests.pop_front();

Not really needed as those objects are cheap to copy, but it could even be:

    auto& request = _requests.front();
    const auto image = std::move(request.image);
    const auto tasks = std::move(request.tasks);
    const auto promise = std::move(request.promise); 
    _requests.pop_front();

deflect/StreamSendWorker.cpp
```diff
@@ -68,40 +68,43 @@ void StreamSendWorker::_run()
...
         }
 
-        if (request.tasks & Request::TASK_FINISH)
+        if (tasks & Request::TASK_FINISH)
         {
             if (!_stream.sendFinish())
             {
-                request.promise->set_value(false);
+                promise->set_value(false);
                 _requests.pop_back();
```
rdumusc: I think those two pop_back() should no longer be here!

rdumusc (commented):
Good otherwise

deflect/StreamSendWorker.cpp
eile: done

deflect/StreamSendWorker.cpp
eile: done

eile: good catch

rdumusc (approved):

eile:
@rdumusc : I'll let you manage the merge of this, Equalizer and Tide.

rdumusc:
@eile I'll handle Tide later today, but it would be easier if you could just update the sha1 of Deflect in the Equalizer PR and merge

eile:
Will do

PR #156, opened 2017-03-16
Deflect can be released as a .deb package.

rdumusc:
The installed DesktopStreamer application can now be launched from any standard Linux desktop app menu or launcher.


rdumusc:
still needs CMake update: https://github.com/Eyescale/CMake/pull/545

CMakeLists.txt
```diff
@@ -48,12 +48,12 @@ if(Boost_FOUND)
 endif()
 add_subdirectory(doc)
 
+set(CPACK_PACKAGE_VENDOR "bluebrain.epfl.ch")
+set(CPACK_DEBIAN_PACKAGE_DEPENDS "qtbase5-dev, libturbojpeg")
```
hernando: Is this list really complete? Seems too short to me.

hernando (approved):
Looks OK

dnachbaur (approved):

CMakeLists.txt
rdumusc: It is, tested on a fresh Ubuntu 16.04 VM :-)
Only simplestreamer uses glut, but since it is just a dummy test application I don't think it's worth including.

PR #157, opened 2017-03-21
Server: improved even registration mechanism

rdumusc:
The server now relies on a std::promise to check the success of a request to register for events instead of passively listening for a QSignal to come back. This change brings two improvements:

- It fixes an issue with multiple clients, where all the clients were being notified when any of them tried to register, causing a warning message to be printed.
- The Stream client is notified also if the request is not handled at all by the Server application (broken promise).

### DIFF
--- a/deflect/ServerWorker.cpp
+++ b/deflect/ServerWorker.cpp
@@ -117,16 +117,6 @@ void ServerWorker::closeConnection(const QString uri)
     emit(connectionClosed());
 }
 
-void ServerWorker::replyToEventRegistration(const QString uri,
-                                            const bool success)
-{
-    if (uri != _streamId)
-        return;
-
-    _registeredToEvents = success;
-    _sendBindReply(_registeredToEvents);
-}
-
 void ServerWorker::_processMessages()
 {
     const qint64 headerSize(MessageHeader::serializedSize);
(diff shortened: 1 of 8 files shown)


hernando (commented):
I don't feel qualified to judge this change. You will need to explain it to me in person.

deflect/ServerWorker.cpp
```diff
@@ -268,7 +258,18 @@ void ServerWorker::_handleMessage(const MessageHeader& messageHeader,
...
-            emit registerToEvents(_streamId, exclusive, this);
+            auto promise = std::make_shared<std::promise<bool>>();
+            auto future = promise->get_future();
+            emit registerToEvents(_streamId, exclusive, this,
+                                  std::move(promise));
+            try
+            {
+                _registeredToEvents = future.get();
+            }
+            catch (...)
```
hernando, line 269: catch broken promise instead of anything?

hernando (approved):

deflect/ServerWorker.cpp
rdumusc: I find it safer to catch everything in case the promise holds an exception instead of a value. In all cases, it is reasonable to assume that if an exception happens then the registration must have failed.

PR #160, opened 2017-04-13
Hide QSocketNotifer false warnings issued by the async Stream API

rdumusc:


apps/DesktopStreamer/main.cpp
```diff
@@ -40,14 +40,16 @@
...
 
+#include <deflect/utils.h>
 #include <deflect/version.h>
 
 #include <QCommandLineParser>
 
 int main(int argc, char* argv[])
 {
-    QApplication app(argc, argv);
+    deflect::blockFalseWarnings();
```
eile, line 50: The application should not see this. Have a static initializer in the method which emits this warning?

eile (changes requested):

apps/DesktopStreamer/main.cpp
rdumusc: Maybe... I wasn't too sure about this. If we place it somewhere in Stream (e.g. constructor) it will overwrite a previously set qInstallMessageHandler() and there is no getter to check beforehand. But it seems that the setter returns the previous handler so we could try to set the new one and if the returned value is not nullptr set back the previous one. A bit fishy but I can try.

rdumusc: Sorry, the default QtMessageHandler returned is NOT nullptr so it's impossible to check whether there is a handler already registered or not. Each application has to do it explicitly.

rdumusc:
Superseeded by https://github.com/BlueBrain/Deflect/pull/162

PR #163, opened 2017-04-21
Stereo streaming accepts side-by-side images as input

rdumusc:
This is useful for applications like Brayns which render natively in
side-by-side format. Splitting the image into a left and right part
during the tiling process is more optimal and user-friendly than
asking the application to do it and make two send() calls.

### DIFF
--- a/deflect/ImageSegmenter.cpp
+++ b/deflect/ImageSegmenter.cpp
@@ -1,5 +1,5 @@
 /*********************************************************************/
-/* Copyright (c) 2013-2016, EPFL/Blue Brain Project                  */
+/* Copyright (c) 2013-2017, EPFL/Blue Brain Project                  */
 /*                          Raphael Dumusc <raphael.dumusc@epfl.ch>  */
 /*                          Stefan.Eilemann@epfl.ch                  */
 /* All rights reserved.                                              */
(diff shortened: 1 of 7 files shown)


deflect/ImageSegmenter.cpp
```diff
@@ -163,113 +173,92 @@ bool ImageSegmenter::_generateRaw(const ImageWrapper& image,
...
+    if (image.view == View::side_by_side)
+    {
+        if (image.width % 2 != 0)
+            throw std::runtime_error("side_by_side image width must be even!");
 
-    unsigned int uniformSegmentWidth = image.width;
-    unsigned int uniformSegmentHeight = image.height;
+        // create copy of segments for right view
+        auto segmentsRight = segments;
+        for (auto&& segment : segmentsRight)
```
dnachbaur: it's more a 'real' reference here, no? modifying a temporary at least looks wrong.

dnachbaur (approved):

deflect/ImageSegmenter.cpp
rdumusc: Yeah, in this case auto&& it's not an rvalue reference but a universal reference... I amended it to avoid confusion. Using auto& or auto&& make no difference here, but I let you google about it.

PR #174, opened 2017-07-03
Performance improvements for sending small images (tested with 64x64)

dnachbaur:
- directly compress them in the call thread
- keep JPEG out buffer allocated to save one memory allocation

### DIFF
--- a/deflect/ImageJpegCompressor.h
+++ b/deflect/ImageJpegCompressor.h
@@ -71,6 +71,7 @@ public:
 
 private:
     tjhandle _tjHandle;
+    std::vector<unsigned char> _tjJpegBuf;
 };
 }
 
--- a/deflect/StreamSendWorker.cpp
+++ b/deflect/StreamSendWorker.cpp
@@ -49,6 +49,7 @@
 namespace
 {
 const unsigned int SEGMENT_SIZE = 512;
+const unsigned int SMALL_IMAGE_SIZE = 64;
 }
 
 namespace deflect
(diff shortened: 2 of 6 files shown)


deflect/ImageJpegCompressor.h
```diff
@@ -71,6 +71,7 @@ class ImageJpegCompressor
 
 private:
     tjhandle _tjHandle;
+    std::vector<unsigned char> tjJpegBuf;
```
rdumusc: _tjJpegBuf

deflect/StreamSendWorker.cpp
```diff
@@ -128,6 +128,17 @@ Stream::Future StreamSendWorker::enqueueImage(const ImageWrapper& image,
         return promise.get_future();
     }
 
+    if (image.width <= 64 && image.height <= 64 &&
```
rdumusc: #define 64 at the top

deflect/StreamSendWorker.cpp
```diff
@@ -128,6 +128,17 @@ Stream::Future StreamSendWorker::enqueueImage(const ImageWrapper& image,
...
+        image.compressionPolicy == COMPRESSION_ON)
+    {
+        auto segment = _imageSegmenter.compressSingleSegment(image);
+        auto tasks = std::vector<Task>{
+            [this, segment] { return _sendSegment(segment); }};
+        if (finish)
+            tasks.emplace_back([this] { return _sendFinish(); });
+        return _enqueueRequest(std::move(tasks));
```
rdumusc: these last 3 lines are the same as the normal case below, reuse them with an if/else for setting tasks?

rdumusc (commented):
Good, a few minor details!

dnachbaur:
updated

rdumusc (approved):

PR #175, opened 2017-07-28
Add deflect::Observer which can be used to only receive events w/o the need to send images

dnachbaur:


deflect/Observer.h
```diff
@@ -0,0 +1,199 @@
...
+#include <functional>
+#include <memory>
+#include <string>
+
+namespace deflect
+{
+class StreamPrivate;
+
+/**
+ */
```
rdumusc: empty docstring

adjust documentation in this file a bit, explaining how to use an Observer and replacing "stream" with "observer" where applicable

deflect/Server.cpp
```diff
@@ -55,11 +55,17 @@ const int Server::defaultPortNumber = DEFAULT_PORT_NUMBER;
...
 {
 public:
-    FrameDispatcher frameDispatcher;
+    Impl(QObject* parent)
+        : frameDispatcher(
+              new FrameDispatcher(parent)) // we be deleted by parent
+    {
+    }
+
+    FrameDispatcher* frameDispatcher;
```
rdumusc: typo we be deleted

deflect/StreamSendWorker.cpp
```diff
@@ -164,6 +164,20 @@ Stream::Future StreamSendWorker::enqueueClose()
...
+    return _enqueueRequest({[this] {
+        return _send(MESSAGE_TYPE_OBSERVER_OPEN,
+                     QByteArray::number(NETWORK_PROTOCOL_VERSION));
+    }});
+}
+
+Stream::Future StreamSendWorker::enqueueObserverClose()
+{
+    return _enqueueRequest(
+        {[this] { return _send(MESSAGE_TYPE_OBSERVER_QUIT, {}); }});
```
rdumusc: do we need a separate MESSAGE_TYPE_OBSERVER_QUIT? Couldn't we just use the normal MESSAGE_TYPE_QUIT?

deflect/ServerWorker.cpp
```diff
@@ -86,7 +86,10 @@ ServerWorker::~ServerWorker()
     // if other senders are still active / resp. the window gets closed if no
     // more senders contribute to it.
     if (!_streamId.isEmpty())
+    {
         emit removeStreamSource(_streamId, _sourceId);
+        emit removeObserver(_streamId, _sourceId);
```
rdumusc: look like a bit brute force, maybe remember the stream type and emit just the correct signal?

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 #                          Daniel Nachbaur <daniel.nachbaur@epfl.ch>
 
 cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
-project(Deflect VERSION 0.13.1)
+project(Deflect VERSION 0.14.0)
```
rdumusc: I suspect the split of Stream in Observer + Stream to break the ABI as well, no?

dnachbaur: probably yes. Will bump for safety.

deflect/StreamSendWorker.cpp
dnachbaur: Couldn't find a satisfying solution to make that work. Now it's clearer having two messages indicating the the source is a observer.

deflect/ServerWorker.cpp
dnachbaur: hmm, if the type is remembered, the upper comment could also be resolved. I'll have a look

rdumusc (approved):
The last thing I don't know is if we should update the @version tags in Observer.h, now that they are in a new class. Both updating and keeping them as is seems a bit wrong.

dnachbaur:
I'm actually wondering what the version tag corresponds to the CMakeLists.txt version. 0.13 == 1.3? But there's a 1.5 version tag in Observer.h as well...
Maybe mention that this class is new since version 1.5, but for the methods we keep the version we had as it didn't break looking from the Stream's point-of-view.

PR #176, opened 2017-08-03
OPT: Lock-free request queueing for multi-threaded stream clients (e.g. KNL)

dnachbaur:


deflect/StreamSendWorker.cpp
```diff
@@ -136,6 +170,13 @@ Stream::Future StreamSendWorker::enqueueImage(const ImageWrapper& image,
     {
         auto segment = _imageSegmenter.compressSingleSegment(image);
         tasks.emplace_back([this, segment] { return _sendSegment(segment); });
+        _requests.enqueue({nullptr, tasks, false});
+
+        // as we expect to encounter a lot of these small sends, be optimistic
+        // and fulfill the promise already to reduce load in the send thread
+        std::promise<bool> promise;
+        promise.set_value(true);
+        return promise.get_future();
```
rdumusc: OK, not sure why though? Is it because promise.set_value(true); takes too much time on KNL?
- move "_requests.enqueue({nullptr, tasks, false});" below the comment
- return make_ready_future(true); ?

tests/cpp/ServerTests.cpp
```diff
@@ -376,3 +376,147 @@ BOOST_AUTO_TEST_CASE(testOneObserverAndOneStream)
...
+    const unsigned int width = 1920;
+    const unsigned int height = 1080;
+
+    struct Segment
+    {
+        unsigned int x;
+        unsigned int y;
+    };
+    std::vector<Segment> segments;
+    for (unsigned int i = 0; i < width + segmentSize - 1; i += segmentSize)
```
rdumusc: Hmmm, these termination condition look complicated, I think it should be just i < width and j < height. Maybe add:
BOOST_REQUIRE_EQUAL(numSegments, std::ceil((float)width / segmentSize) * std::ceil((float)height / segmentSize));
Also, it see that it does not matter too much here, but if you had checked the size of the received frame it's height would not have been correct because (1080 % 64) != 0 (the tile of the last row are too high)

tests/cpp/ServerTests.cpp
```diff
@@ -376,3 +376,147 @@ BOOST_AUTO_TEST_CASE(testOneObserverAndOneStream)
...
+    {
+        for (unsigned int j = 0; j < height + segmentSize - 1; j += segmentSize)
+            segments.emplace_back(Segment{i, j});
+    }
+    const unsigned int numSegments = segments.size();
+
+    const unsigned int byte = segmentSize * segmentSize * 4;
+    std::unique_ptr<uint8_t[]> pixels(new uint8_t[byte]);
+    ::memset(pixels.get(), 0, byte);
```
rdumusc: std::vector<uint8_t> pixels(bytes, 0); ?

rdumusc (approved):
All good!

PR #177, opened 2017-08-07
Various fixes

dnachbaur:
* Throw on failed connection to server in stream ctor; refactor test deflect servers
* Transform error message printing into exceptions
* Catch errors regarding invalid JPEG quality values
* Allow uncompressed small tile sending
* Move sendSizeHints and sendData also to Observer

### DIFF
--- a/deflect/ImageJpegCompressor.cpp
+++ b/deflect/ImageJpegCompressor.cpp
@@ -42,6 +42,7 @@
 #include "ImageWrapper.h"
 
 #include <iostream>
+#include <sstream>
 
 namespace deflect
 {
@@ -72,7 +73,8 @@ int _getTurboJpegFormat(const PixelFormat pixelFormat)
     case ABGR:
         return TJPF_XBGR;
     default:
-        std::cerr << "unknown pixel format" << std::endl;
+        throw std::invalid_argument("unknown pixel format " +
+                                    std::to_string((int)pixelFormat));
         return TJPF_RGB;
     }
 }
@@ -88,7 +90,8 @@ int _getTurboJpegSubsamp(const ChromaSubsampling subsampling)
     case ChromaSubsampling::YUV420:
         return TJSAMP_420;
     default:
-        std::cerr << "unknown subsampling format" << std::endl;
+        throw std::invalid_argument("unknown subsampling format " +
+                                    std::to_string((int)subsampling));
         return TJSAMP_444;
     }
 }
@@ -101,12 +104,8 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
     // pointer to comply with the incorrect API.
     unsigned char* tjSrcBuffer = (unsigned char*)sourceImage.data;
     if (!tjSrcBuffer)
-    {
-        std::cerr
-            << "libjpeg-turbo image conversion failure: source image is NULL"
-            << std::endl;
-        return QByteArray();
-    }
+        throw std::invalid_argument(
+            "libjpeg-turbo image conversion failure: source image is NULL");
 
     tjSrcBuffer +=
         imageRegion.y() * sourceImage.width * sourceImage.getBytesPerPixel();
@@ -132,9 +131,9 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
                           tjJpegQual, tjFlags);
     if (err != 0)
     {
-        std::cerr << "libjpeg-turbo image conversion failure: "
-                  << tjGetErrorStr() << std::endl;
-        return QByteArray();
+        std::stringstream msg;
+        msg << "libjpeg-turbo image conversion failure: " << tjGetErrorStr();
+        throw std::runtime_error(msg.str());
     }
 
     return QByteArray((const char*)ptr, tjJpegSize);
--- a/deflect/ImageJpegCompressor.cpp
+++ b/deflect/ImageJpegCompressor.cpp
@@ -100,6 +100,14 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
     // though it does not modify it. It can "safely" be cast to non-const
     // pointer to comply with the incorrect API.
     unsigned char* tjSrcBuffer = (unsigned char*)sourceImage.data;
+    if (!tjSrcBuffer)
+    {
+        std::cerr
+            << "libjpeg-turbo image conversion failure: source image is NULL"
+            << std::endl;
+        return QByteArray();
+    }
+
     tjSrcBuffer +=
         imageRegion.y() * sourceImage.width * sourceImage.getBytesPerPixel();
     tjSrcBuffer += imageRegion.x() * sourceImage.getBytesPerPixel();
@@ -124,7 +132,8 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
                           tjJpegQual, tjFlags);
     if (err != 0)
     {
-        std::cerr << "libjpeg-turbo image conversion failure" << std::endl;
+        std::cerr << "libjpeg-turbo image conversion failure: "
+                  << tjGetErrorStr() << std::endl;
         return QByteArray();
     }
 
--- a/deflect/ImageJpegCompressor.h
+++ b/deflect/ImageJpegCompressor.h
@@ -65,6 +65,7 @@ public:
      * @param sourceImage The source image containing uncompressed image data.
      * @param imageRegion The region of the image to be compressed. Must not
      *        exceed image dimensions.
+     * @return compressed image if successful, otherwise QByteArray.isEmpty()
      */
     DEFLECT_API QByteArray computeJpeg(const ImageWrapper& sourceImage,
                                        const QRect& imageRegion);
--- a/deflect/ImageJpegCompressor.h
+++ b/deflect/ImageJpegCompressor.h
@@ -65,7 +65,9 @@ public:
      * @param sourceImage The source image containing uncompressed image data.
      * @param imageRegion The region of the image to be compressed. Must not
      *        exceed image dimensions.
-     * @return compressed image if successful, otherwise QByteArray.isEmpty()
+     * @return compressed image
+     * @throw std::invalid_argument if sourceImage.data is nullptr
+     * @throw std::runtime_error if JPEG compression failed
      */
     DEFLECT_API QByteArray computeJpeg(const ImageWrapper& sourceImage,
                                        const QRect& imageRegion);
--- a/deflect/ImageSegmenter.cpp
+++ b/deflect/ImageSegmenter.cpp
@@ -69,19 +69,33 @@ bool ImageSegmenter::generate(const ImageWrapper& image, const Handler& handler)
     return _generateRaw(image, handler);
 }
 
-Segment ImageSegmenter::compressSingleSegment(const ImageWrapper& image)
+Segment ImageSegmenter::createSingleSegment(const ImageWrapper& image)
 {
-#ifdef DEFLECT_USE_LIBJPEGTURBO
     auto segments = _generateSegments(image);
     if (segments.size() > 1)
         throw std::runtime_error(
-            "compressSingleSegment only works for small images");
-    _computeJpeg(segments[0], false);
-    return segments[0];
+            "createSingleSegment only works for small images");
+
+    auto& segment = segments[0];
+
+    if (image.compressionPolicy == COMPRESSION_OFF)
+    {
+        segment.imageData.reserve(segment.parameters.width *
+                                  segment.parameters.height *
+                                  image.getBytesPerPixel());
+        segment.parameters.dataType = DataType::rgba;
+        segment.imageData.append((const char*)image.data,
+                                 int(image.getBufferSize()));
+    }
+    else
+#ifdef DEFLECT_USE_LIBJPEGTURBO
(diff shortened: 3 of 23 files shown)


deflect/ImageSegmenter.cpp
```diff
@@ -104,8 +104,20 @@ bool ImageSegmenter::_generateJpeg(const ImageWrapper& image,
...
-    QtConcurrent::map(segments, std::bind(&ImageSegmenter::_computeJpeg, this,
-                                          std::placeholders::_1, true));
+    auto future =
+        QtConcurrent::map(segments,
+                          std::bind(&ImageSegmenter::_computeJpeg, this,
+                                    std::placeholders::_1, true));
+
+    try
+    {
+        future.waitForFinished();
```
rdumusc: No, we shouldn't wait for the result. The sending should start while the compression is in progress (see comments below).

deflect/ImageJpegCompressor.cpp
```diff
@@ -124,8 +125,10 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
                           tjJpegQual, tjFlags);
     if (err != 0)
     {
-        std::cerr << "libjpeg-turbo image conversion failure" << std::endl;
-        return QByteArray();
+        std::stringstream msg;
+        msg << "libjpeg-turbo image conversion failure: " << tjGetErrorStr()
+            << std::endl;
+        throw std::runtime_error(msg.str());
```
rdumusc: OK but the header should mention that the function can throw.

rdumusc (commented):
Good in general!

deflect/ImageSegmenter.cpp
dnachbaur: Of course. I just focused on getting the exception from the async execution. Too bad

dnachbaur:
I kept the empty bytearray for indicating errors and check it accordingly.

tests/cpp/StreamTests.cpp
```diff
@@ -116,3 +116,49 @@ BOOST_AUTO_TEST_CASE(testWhenNoSteamIdProvidedThenARandomOneIsGenerated)
         qunsetenv(STREAM_HOST_ENV_VAR);
     }
 }
+
+// Note: the following send tests only work as the small segment send does not
+// need the sendWorker thread to be present. So fortunate for us we can skip
+// setting up a stream server.
+BOOST_AUTO_TEST_CASE(testErrorOnUncompressedRGBA)
```
rdumusc: Great, this was indeed missing! Let's be fully complete and split this in three test cases:
- testSendUncompressedRGBA
- testErrorOnUnsupportedUncompressedFormats
- testErrorOnNullUncompressedImage

tests/cpp/ServerTests.cpp
```diff
@@ -526,3 +526,71 @@ BOOST_AUTO_TEST_CASE(testThreadedSmallSegmentStream)
...
+    {
+        deflect::Stream stream(testStreamId.toStdString(), "localhost",
+                               server->serverPort());
+        BOOST_REQUIRE(stream.isConnected());
+
+        processServerMessages();
+
+        deflect::ImageWrapper bigImage(nullptr, 1000, 1000, deflect::ARGB);
+        bigImage.compressionPolicy = deflect::COMPRESSION_ON;
+        BOOST_CHECK(!stream.send(bigImage).get());
```
rdumusc: I that the only line we are truly testing? The amount of boilerplate code in this test suite is starting to be worrying, we should look for a way to mock the server / connection or at least reuse a shared Fixture.

deflect/ImageJpegCompressor.h
```diff
@@ -65,6 +65,7 @@ class ImageJpegCompressor
      * @param sourceImage The source image containing uncompressed image data.
      * @param imageRegion The region of the image to be compressed. Must not
      *        exceed image dimensions.
+     * @return compressed image if successful, otherwise QByteArray.isEmpty()
```
rdumusc: +1, this was missing

deflect/ImageJpegCompressor.cpp
```diff
@@ -132,9 +131,9 @@ QByteArray ImageJpegCompressor::computeJpeg(const ImageWrapper& sourceImage,
                           tjJpegQual, tjFlags);
     if (err != 0)
     {
-        std::cerr << "libjpeg-turbo image conversion failure: "
-                  << tjGetErrorStr() << std::endl;
-        return QByteArray();
+        std::stringstream msg;
+        msg << "libjpeg-turbo image conversion failure: " << tjGetErrorStr();
+        throw std::invalid_argument(msg.str());
```
rdumusc: I would say this is more a runtime_error at this point

deflect/ImageJpegCompressor.h
```diff
@@ -65,7 +65,9 @@ class ImageJpegCompressor
      * @param sourceImage The source image containing uncompressed image data.
      * @param imageRegion The region of the image to be compressed. Must not
      *        exceed image dimensions.
-     * @return compressed image if successful, otherwise QByteArray.isEmpty()
+     * @return compressed image
+     * @throw std::invalid_argument if sourceImage.data is nullptr or
+     *                              tjCompress2 failed
```
rdumusc: don't mention tjCompress2 in the doc, it's an implementation detail and has the potential to become outdated

deflect/ImageSegmenter.h
```diff
@@ -98,6 +98,8 @@ class ImageSegmenter
      *
      * @param image The image to be compressed
      * @return the compressed segment
+     * @throw std::runtime_error if image is too big
```
rdumusc: too big -> invalid_argument?

deflect/StreamSendWorker.cpp
```diff
@@ -117,14 +122,25 @@ void StreamSendWorker::run()
...
+                try
+                {
+                    if (!task())
+                    {
+                        if (request.promise)
+                            request.promise->set_value(false);
+                        break;
+                    }
+                    if (request.promise)
+                        request.promise->set_value(true);
```
rdumusc: I don't think this works, does it pass unit tests? you have changed the logic and doing set_value() for each task, which should fail at the second set (and be caught again by the catch (...)). Recommendation: Keep the logic as it was before, with the try/catch surrounding the entire for-loop.

deflect/StreamSendWorker.cpp
```diff
@@ -151,44 +167,54 @@ Stream::Future StreamSendWorker::enqueueImage(const ImageWrapper& image,
...
         if (image.compressionQuality < 1 || image.compressionQuality > 100)
         {
-            std::cerr
-                << "JPEG compression quality must be between 1 and 100, got "
+            std::stringstream msg;
+            msg << "JPEG compression quality must be between 1 and 100, got "
                 << image.compressionQuality << std::endl;
-            return make_ready_future(false);
+            return make_exception_future<bool>(
+                std::make_exception_ptr(std::invalid_argument(msg.str())));
```
rdumusc: move std::make_exception_ptr inside make_exception_future to avoid repeating it?

tests/cpp/StreamTests.cpp
```diff
@@ -142,7 +142,23 @@ BOOST_AUTO_TEST_CASE(testErrorOnUnsupportedUncompressedFormats)
...
+        BOOST_CHECK_THROW(stream.send(image).get(), std::invalid_argument);
+    }
+}
+
+BOOST_AUTO_TEST_CASE(testSuccessOnCompressedFormats)
+{
+    deflect::Stream stream("id", "dummyhost");
+    std::vector<unsigned char> pixels(4 * 4 * 4);
+
+    std::vector<deflect::PixelFormat> unsupportedUncompressedFormats = {
```
rdumusc: name it allFormats here? I think you can even use auto for the variable type (init list).

tests/cpp/StreamTests.cpp
```diff
@@ -53,66 +56,66 @@ const char* STREAM_ID_ENV_VAR = "DEFLECT_ID";
...
-BOOST_AUTO_TEST_CASE(testParameterizedConstructorWithValues)
+BOOST_AUTO_TEST_CASE(testConstructionWithNoServer)
 {
-    const deflect::Stream stream("mystream", "somehost");
-    BOOST_CHECK_EQUAL(stream.getId(), "mystream");
-    BOOST_CHECK_EQUAL(stream.getHost(), "somehost");
+    BOOST_CHECK_THROW(deflect::Stream("id", "localhost"), std::runtime_error);
 }
 
-BOOST_AUTO_TEST_CASE(testDefaultConstructorReadsEnvironmentVariables)
```
rdumusc: this test case got lost, but I still valid

tests/cpp/ServerTests.cpp
```diff
@@ -510,87 +369,31 @@ BOOST_AUTO_TEST_CASE(testThreadedSmallSegmentStream)
...
     }
 
     // handle close of streamer
-    processServerMessages();
-
-    serverThread.quit();
-    serverThread.wait();
+    processServerMessage();
 
-    BOOST_CHECK_EQUAL(openedStreams, 0);
```
rdumusc: no longer checked?

rdumusc (commented):
This improves the readability quite a bit, good job!

tests/cpp/StreamTests.cpp
dnachbaur: It cannot be replicated if a server is used that does not run on port 1701, which shouldn't be hardcoded in a unit test. DEFLECT_PORT as an env variable can be the only solution then.

deflect/StreamSendWorker.cpp
dnachbaur: There is ```make_exception_future<bool>(std::current_exception());``` in StreamSendWorker which wouldn't work then

tests/cpp/ServerTests.cpp
dnachbaur: I wasn't fully done yet yesterday, you reviewed too early :) I have not adapted all tests in here yet

deflect/StreamSendWorker.cpp
rdumusc: maybe we could have two, or an overload for Exception=std::exception_ptr, something like:
```
template <typename T>
std::future<T> make_exception_future(std::exception_ptr&& e)
{
    std::promise<T> promise;
    promise.set_exception(std::move(e));
    return promise.get_future();
}
template <typename T, typename Exception>
std::future<T> make_exception_future(Exception&& e)
{
    return make_exception_future(std::make_exception_ptr(std::move(e)));
}


```

dnachbaur: That would work. Looks nice, thx!

tests/cpp/ServerTests.cpp
dnachbaur: should be checked again

deflect/Observer.h
```diff
@@ -69,16 +69,21 @@ class StreamPrivate;
 class Observer
 {
 public:
+    /** The default communication port */
+    static const unsigned short defaultPortNumber;
```
rdumusc, line 73: use this constant also in the 2 constructors with params (Observer + Stream)

deflect/StreamSendWorker.cpp
```diff
@@ -167,16 +167,15 @@ Stream::Future StreamSendWorker::enqueueImage(const ImageWrapper& image,
 {
     if (_pendingFinish)
     {
-        return make_exception_future<bool>(std::make_exception_ptr(
-            std::invalid_argument("Pending finish, no send allowed")));
+        return make_exception_future<bool>(
+            std::invalid_argument("Pending finish, no send allowed"));
```
rdumusc: std::runtime_error here? Does not depend on the arguments

tests/cpp/StreamTests.cpp
rdumusc: ping?

tests/mock/DeflectServer.cpp
```diff
@@ -0,0 +1,142 @@
...
+                     });
+}
+
+DeflectServer::~DeflectServer()
+{
+    _thread.quit();
+    _thread.wait();
+}
+
+void DeflectServer::processServerMessage()
```
rdumusc: currently this is waitForMessage(), no processing happens here

tests/mock/MinimalDeflectServer.h
```diff
@@ -0,0 +1,59 @@
...
+/*********************************************************************/
+
+#ifndef DEFLECT_MINIMAL_DEFLECTSERVER_H
+#define DEFLECT_MINIMAL_DEFLECTSERVER_H
+
+#include "MockServer.h"
+
+#include <QThread>
+
+class MinimalDeflectServer
```
rdumusc, line 47: I was going to suggest letting DeflectServer derive from MinimalDeflectServer, but realized they are not the same, because this one is in fact a wrapper for the MockServer. We should find clearer names...

dnachbaur: No, we should treat as good now and merge this :)

rdumusc: OK :-)

rdumusc (approved):

deflect/Observer.h
dnachbaur: Done

deflect/StreamSendWorker.cpp
dnachbaur: Done

PR #178, opened 2017-08-10
Fix missing sendFrame after re-adding streams

dnachbaur:


dnachbaur:
Retest this please

rdumusc (commented):
Hmmm no, I think pixelStreamOpened(uri); should really be emitted only once. I would say the problem is more with the fact that the frameDispatcher ignores the first requestFrame if there is no Stream connected.

deflect/FrameDispatcher.cpp
```diff
@@ -172,10 +172,11 @@ void FrameDispatcher::deleteStream(const QString uri)
     if (_impl->streamBuffers[uri].getSourceCount() == 0 &&
         _impl->streamBuffers.count(uri))
     {
-        _impl->streamBuffers.erase(uri);
-
         if (_impl->observers[uri] == 0)
+        {
```
rdumusc: move in other if

tests/cpp/ServerTests.cpp
```diff
@@ -196,7 +196,127 @@ BOOST_AUTO_TEST_CASE(testOneObserverAndOneStream)
...
+                                   serverPort());
+            BOOST_REQUIRE(stream.isConnected());
+        }
+    }
+
+    waitForMessage();
+    BOOST_CHECK_EQUAL(getOpenedStreams(), 0);
+}
+
+BOOST_AUTO_TEST_CASE(closeAndReopenStreamWithObserverStayingAlive)
```
rdumusc: maybe not needed

rdumusc (approved):

PR #180, opened 2017-08-30
DesktopStreamer: fix uncaught exception after #177

rdumusc:
Also report stream connection errors in the GUI instead of cerr.

PR #181, opened 2017-08-30
Fixed unit test deadlock due to incorrect execption handling

rdumusc:
This bug was introduced in #177 and could result in the global QThreadPool threads being locked forever.


dnachbaur (approved):

rdumusc:
damn, still not fixed..!

rdumusc:
retest this please

rdumusc:
I'm merging because it already improves the situation (and #180 refuses to build without this). But the previous failure should not be ignored: https://bbpcode.epfl.ch/ci/job/oss.Deflect.github/405/build_type=Release,platform=cscsviz-vm/console

rdumusc:
for the record, the remaining test failures are mostly due to concurrent calls to BOOST_CHECK

PR #183, opened 2017-09-12
Linux travis

eile:


tests/cpp/ServerTests.cpp
```diff
@@ -116,7 +116,7 @@ BOOST_AUTO_TEST_CASE(testDataReceivedByServer)
     bool received = false;
     setDataReceivedCallback([&](const QString id, QByteArray data) {
         SAFE_BOOST_CHECK_EQUAL(id.toStdString(), testStreamId.toStdString());
-        SAFE_BOOST_CHECK_EQUAL(data.toStdString(), sentData);
+        SAFE_BOOST_CHECK_EQUAL(std::string(data.constData()), sentData);
```
rdumusc, line 119: why changed?

eile: .toStdString did not exist in Qt 5.2 which is installed on trusty

rdumusc: Ouch, 5.2 on Trusty? This is going to be problematic to support in the long run. We removed our internal 14.04 builds a year ago and 18.04 will soon be here. Maybe it's OK to do the effort for some low-level libs, but we can't do it for bigger projects. Tide for example requires Qt 5.4 minimum.

rdumusc (approved):

tests/cpp/ServerTests.cpp
eile: Yes, I was surprised too. Let's hope the offer xenial soon. For now this is only Deflect as a dependency for Equalizer.

PR #195, opened 2018-05-02
Deflect 1.0

rdumusc:


deflect/server/Tile.h
```diff
@@ -0,0 +1,85 @@
...
+    //@{
+    uint32_t x = 0u; /**< The x position in pixels. */
+    uint32_t y = 0u; /**< The y position in pixels. */
+    //@}
+
+    /** @name Dimensions */
+    //@{
+    uint32_t width = 0u;  /**< The width in pixels. */
+    uint32_t height = 0u; /**< The height in pixels. */
+    //@}
```
dnachbaur, line 66: those four and the data type is a repetition from the segment parameters. reuse them here?

deflect/types.h
```diff
@@ -71,6 +71,16 @@ enum class RowOrder
     bottom_up /**< OpenGL image with (0,0) at the bottom-left corner. */
 };
 
+/** The possible formats for image data. */
+enum class Format : std::uint8_t
+{
+    rgba = 0, // equivalent to old compressed=false property
```
dnachbaur: with 1.0, maybe those references to old things are no longer needed to be mentioned

deflect/server/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  Frame.cpp
+  FrameDispatcher.cpp
+  Server.cpp
+  ServerWorker.cpp
+  ReceiveBuffer.cpp
+  SourceBuffer.cpp
+)
+
+set(DEFLECTSERVER_LINK_LIBRARIES
+  PUBLIC Deflect Qt5::Core Qt5::Network
```
dnachbaur: network is public?

deflect/types.h
```diff
@@ -103,24 +103,29 @@ std::future<T> make_exception_future(Exception&& e)
...
 struct Event;
 struct ImageWrapper;
-struct MessageHeader;
 struct Segment;
 struct SegmentParameters;
 struct SizeHints;
 
+using Segments = std::vector<Segment>;
+
+namespace server
```
dnachbaur: move them to a separate server/types.h?

doc/Changelog.md
```diff
@@ -6,7 +6,8 @@ Changelog {#Changelog}
 ### 1.0.0 (git master)
 * [194](https://github.com/BlueBrain/Deflect/pull/194):
   Reset API versionning to 1.0, remove deprecated functions (Stream::asyncSend,
-  ImageWrapper::swapYAxis).
+  ImageWrapper::swapYAxis). Moved all server-specific classes to a separate
```
dnachbaur: that change is not part of 194?!

CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
 
-# Copyright (c) 2013-2017, EPFL/Blue Brain Project
+# Copyright (c) 2013-2018, EPFL/Blue Brain Project
 #                          Raphael Dumusc <raphael.dumusc@epfl.ch>
 #                          Daniel Nachbaur <daniel.nachbaur@epfl.ch>
 
 cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
-project(Deflect VERSION 0.14.1)
+project(Deflect VERSION 1.0.0)
```
dnachbaur, line 7: yeah!

deflect/StreamEventThread.cpp
```diff
@@ -1,200 +0,0 @@
-/*********************************************************************/
```
dnachbaur, line 1: ? haven't seen this file ever? wow

doc/Changelog.md
```diff
@@ -1,6 +1,13 @@
 Changelog {#Changelog}
 ============
 
+## Deflect 1.0
+
+### 1.0.0 (git master)
+* [194](https://github.com/BlueBrain/Deflect/pull/194):
+  Reset API versionning to 1.0, remove deprecated functions (Stream::asyncSend,
```
dnachbaur: one 'n' too much in versioning

deflect/types.h
rdumusc: You're right, it probably doesn't serve a purpose anymore. Anyone changing this enum will break the network protocol, but that's true of all the other enums as well, so no reason to mark those two as "special".

deflect/server/CMakeLists.txt
rdumusc: Because Server : public QTcpServer. It probably wouldn't be too hard to hide it in the future.

deflect/types.h
rdumusc: OK, will do, also for qt/types.h then.

CMakeLists.txt
rdumusc: :-)

deflect/StreamEventThread.cpp
rdumusc: I found it by accident, it had been forgotten since 2013...!

doc/Changelog.md
rdumusc: thanks, done

deflect/server/Tile.h
rdumusc: This is a tricky point. The existing structs Segment and SegementParameters are shared between Deflect and DeflectServer because they are part of the networking protocol. However, they were also part of the public DeflectServer API because they were exposed through the Frame object. The problem was that DeflectServer depends on Deflect, so the shared classes should logically be in Deflect, but the headers have no reason to be public in Deflect because they are not used by client code. In the end I decided to keep the legacy structs private in Deflect and introduce this new Tile struct for use in Frame in DeflectServer that combines the two. It's nicer to use without the nested SegementParameters object and the public API can evolve separately from the networking API, at the cost of some duplication.

rdumusc:
Damn, two unit tests just failed in "closeAndReopenStreamWithObserverStayingAliveAndRendering" with a failed assert(!frame->tiles.empty()); but I can't reproduce it locally, must be a nasty timing issue...

rdumusc:
retest this please

rdumusc:
retest this please

dnachbaur (approved):

PR #196, opened 2018-05-17
Multi-channel streaming

rdumusc:


doc/Changelog.md
```diff
@@ -4,7 +4,9 @@ Changelog {#Changelog}
 ## Deflect 1.0
 
 ### 1.0.0 (git master)
-* [194](https://github.com/BlueBrain/Deflect/pull/194):
+* [196](https://github.com/BlueBrain/Deflect/pull/196):
```
dnachbaur, line 7: 1.0 is not released yet?

dnachbaur (approved):

doc/Changelog.md
rdumusc: no, it's the current working branch. Do you think we should release it? There wasn't much difference with 0.14.1 feature-wise. With the support for multi-channel it could be worth it.

dnachbaur: Well, the API has changed, like the equalizer build is broken :) Unless you expect more API changes, a release seems fine to me.

PR #198, opened 2018-05-22
Fix unreliable server-side stream close, cleanup FrameDispatcher

rdumusc:


dnachbaur (approved):
Looks still good to me :)

Maybe still test with Tide if it does the job.

rdumusc:
OK thanks, yes, I'm going to deploy that now

PR #200, opened 2018-05-24
Fix data races for ServerTests

dnachbaur:


rdumusc (approved):
very good job, thanks|

PR #203, opened 2018-10-29
Respect size hints when resizing Qml streamer

rdumusc:


rdumusc:
retest this please

deflect/qt/QmlStreamerImpl.cpp
```diff
@@ -246,6 +246,35 @@ void QmlStreamer::Impl::_setupSizeHintsConnections()
...
+{
+    return !size.isEmpty() &&
+           _isWithinSizeHintsRange(static_cast<uint>(size.width()),
+                                   static_cast<uint>(size.height()));
+}
+
+bool QmlStreamer::Impl::_isWithinSizeHintsRange(const uint width,
+                                                const uint height) const
+{
+    const auto minWidth = _sizeHints.minWidth;
```
karjonas:
```suggestion
    const auto minWidth = _sizeHints.minWidth == SizeHints::UNSPECIFIED_SIZE
                              ? 0
                              : _sizeHints.minWidth;
    const auto maxWidth = _sizeHints.maxWidth == SizeHints::UNSPECIFIED_SIZE
                              ? std::numeric_limits<uint>()
                              : _sizeHints.maxWidth;
    const auto minHeight = _sizeHints.minHeight == SizeHints::UNSPECIFIED_SIZE
                               ? 0
                               : _sizeHints.minHeight;
    const auto maxHeight = _sizeHints.maxHeight == SizeHints::UNSPECIFIED_SIZE
                               ? std::numeric_limits<uint>()
                               : _sizeHints.maxHeight;

    return minWidth <= width && width <= maxWidth && minHeight <= height &&
           height <= maxHeight;
```

Do you think this is easier to understand?

rdumusc: Doesn't make a big difference to me, but I can apply your suggestion. Maybe have having a single return point is better style?

karjonas: Having everything constant and one return is a better style imo. But it does look a bit repeated. Anyhow, no big deal.

karjonas (approved):

deflect/qt/QmlStreamerImpl.cpp
rdumusc: This got me thinking and I started looking if there was a good way to check if a value is withing a range and found (https://stackoverflow.com/questions/9255887/) the following option:
```
template <typename T>
struct Interval
{
    T low;
    T high;
    bool contains(const T value) const { return low <= value && value <= high; }
};
template <typename T>
Interval<T> interval(const T low, const T high)
{
    return Interval<T>{low, high};
}
// Then you can be more explicit about what you mean:
if (interval(a, b).contains(value))
    // ...
```
Might be an overkill for this use case but I like how clean this looks... :-)

PR #207, opened 2018-11-20
Fix wrong vertical frame mirror for different sized channels

dnachbaur:


deflect/server/FrameDispatcher.cpp
```diff
@@ -75,9 +76,15 @@ class FrameDispatcher::Impl
 
     void mirrorTilesPositionsVertically(Frame& frame) const
     {
-        const auto height = frame.computeDimensions().height();
+        std::unordered_map<int, int> heights;
+        for (const auto& tile : frame.tiles)
+            heights[tile.channel] = 0;
+
+        for (auto& entry : heights)
+            entry.second = frame.computeDimensions(entry.first).height();
```
rdumusc: How about merging the multiple loops above into one by adding a new function:
```
std::map<uint8_t, QSize> Frame::computeDimensions() const
{
    std::map<uint8_t, QSize> sizes;
    for (const auto& tile : tiles)
    {
        auto& size = sizes[tile.channel];
        size.setWidth(std::max(size.width(), (int)(tile.width + tile.x)));
        size.setHeight(std::max(size.height(), (int)(tile.height + tile.y)));
    }
    return sizes;
}
```

dnachbaur:
Updated

rdumusc (approved):

PR #209, opened 2019-03-04
Deflect server optimizations

hernando:
According to VTune, QMetaObject::activate from the emit receivedTile inside
ServerWorker::_handleMessage was hotspot #1. Using a direct connection
removes it.
The other optimization is trivial, but it was another significant hotspot.

### DIFF
--- a/deflect/server/Server.cpp
+++ b/deflect/server/Server.cpp
@@ -120,14 +120,17 @@ public:
                     &ServerWorker::closeConnections);
 
             // FrameDispatcher
+            // direct connection to avoid race with receivedTile in early tiles
             connect(worker, &ServerWorker::addStreamSource, frameDispatcher,
-                    &FrameDispatcher::addSource);
+                    &FrameDispatcher::addSource, Qt::DirectConnection);
             connect(frameDispatcher, &FrameDispatcher::sourceRejected, worker,
                     &ServerWorker::closeConnection);
+            // direct connection for performance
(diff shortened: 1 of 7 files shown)


hernando:
I haven't checked yet whether the direct connection is thread safe.

hernando:
Now I did. It looks safe to me, but we can discuss it once you're back.

hernando:
[A test has failed](https://travis-ci.org/BlueBrain/Deflect/jobs/501544515) and it could be a thread related thing because I've tried to reproduce it and I've got random errors in ServerTests, but not that one in particular. I even tried makin the ServerWorker::receivedFrameFinished to FrameDispatcher::processFrameFinished connection direct, but it didn't help.

deflect/server/Server.cpp
```diff
@@ -125,7 +125,7 @@ class Server::Impl : public QTcpServer
             connect(frameDispatcher, &FrameDispatcher::sourceRejected, worker,
                     &ServerWorker::closeConnection);
             connect(worker, &ServerWorker::receivedTile, frameDispatcher,
-                    &FrameDispatcher::processTile);
+                    &FrameDispatcher::processTile, Qt::DirectConnection);
```
dnachbaur, line 130: can't judge about concurrent access that happens now that the tiles are added one a different thread. they're probably save, as the finish of a frame is still queued?

hernando:
I would merge this as is for the moment, it's one step further, even if not all issues are resolved.

dnachbaur (approved):

